PR #8, opened 2015-07-03
Add twitter bootstrap to CakePHP

flangofas:
Updated:
- CakePHP 3.0.8
- jQuery 2.1.4
- Bootstrap 3.3.5
- CakePHP Bootstrap Helpers: https://holt59.github.io/cakephp3-bootstrap3-helpers/


src/Template/Layout/default.ctp
```diff
@@ -1,59 +1,42 @@
-<?php
```
mamchenkov: @flangofas the change here shows not only content change, but also permission change from 644 (rw-r--r--) to 755 (rwx-r-xr-x).  We don't need executable permissions on the file.

src/Controller/AppController.php
```diff
@@ -37,5 +53,6 @@ class AppController extends Controller
     public function initialize()
     {
         $this->loadComponent('Flash');
+        $this->layout = 'default';
```
mamchenkov: @flangofas what's happening if you DON'T do this?  I thought the default layout was pretty much assumed.

src/Template/Layout/default.ctp
flangofas: @mamchenkov done!

src/Controller/AppController.php
flangofas: @mamchenkov good catch. :+1:

mamchenkov:
@flangofas  Excellent work! Thanks :+1:

PR #9, opened 2015-07-03
Bugfix/composer installer

mamchenkov:
Implement the fix from cakephp/app regarding installer console, which breaks with recent composer.

Unit tests are temporarily not working.  Will fix shortly.


flangofas:
:+1:

PR #11, opened 2015-09-07
Sync project with Bootstrap plugin' updates and changes

flangofas:

PR #12, opened 2015-09-07
Workaround - when exceeding the limit of unauthenticated user request…

flangofas:
- when exceeding the limit of unauthenticated user requests to Github API
- This fix will probably fix the errors for https://github.com/QoboLtd/project-template-cakephp/pull/11

This issue is reported on composer's repository
Please see here: https://github.com/composer/composer/issues/1861

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ php:
   - 7.0
 
 before_script:
-  - composer install --dev
+  - composer install --dev --prefer-source
   - phpenv rehash
 
 script:


mamchenkov:
@flangofas looking closer at the output of the composer in the [failed TravisCI build](https://travis-ci.org/QoboLtd/project-template-cakephp/jobs/79230575), I see this:

```
  - Installing qobo/pattern (v1.0.6)
    Downloading: connection...    Failed to download qobo/pattern from dist: Could not authenticate against github.com
    Now trying to download from source
  - Installing qobo/pattern (v1.0.6)
    Cloning eaf13a6d3ecd9631d3b7ca14dc2f9dabd65bdcb9
```

So, to me it looks like composer does a fallback on the source installation if the API fails.  So the question is do we really need this.  The bug is open since 2013, but there is not a lot of activity around it.  Maybe it was solved with the fallback and nobody closed it.

flangofas:
@mamchenkov Yeap. Correct. I will look into other solutions.

PR #13, opened 2015-09-08
Composer selfupdate

flangofas:


.travis.yml
```diff
@@ -8,7 +8,8 @@ php:
   - 5.6
   - 7.0
 
-before_script:
+install:
```
mamchenkov: @flangofas what's the difference between before_script and install?

flangofas: @mamchenkov It gets executed earlier so I wanted to try it. :smile:

flangofas:
Am closing this PR since it doesn't fix the bug of travis.

PR #16, opened 2015-09-08
Replaced phpunit cmd

flangofas:


.travis.yml
```diff
@@ -16,7 +16,7 @@ before_script:
   - phpenv rehash
 
 script:
-  - ./vendor/bin/phpunit
+  - phpunit
```
mamchenkov: @flangofas please don't .. it is not guaranteed that phpunit is installed on the system, or which particular version is installed.  We have phpunit in require-dev, so we control which version we ship with the project.  We should use the same one for tests.

flangofas:
@mamchenkov OK, it is just another test to fix travis errors :smile:

PR #20, opened 2016-01-26
Update CakePHP to 3.1.8 (task #795)

flangofas:


mamchenkov:
@flangofas I've updated the project-template, fixed unit tests, and upgraded CakePHP as part of it in #21 ...

PR #22, opened 2016-01-26
Added latest version of CakeDC Users plugin (task #804)

georgeconstantinou:


composer.json
```diff
@@ -6,7 +6,8 @@
         "cakephp/migrations": "~1.0",
         "cakephp/plugin-installer": "*",
 		"qobo/phake-builder": "~1.0",
-        "holt59/cakephp3-bootstrap-helpers": "dev-master"
+        "holt59/cakephp3-bootstrap-helpers": "dev-master",
+        "cakedc/users": "3.1.*"
```
mamchenkov: @georgeconstantinou it's better to use a "~" or "^" wildcard.  These two will make try to use the latest version, while the "*" will satisfy any version.  For example, if you have "3.1.5" installed and version "3.1.7" is available, then "3.1.*" will ignore it, while both "~" and "^" will try to upgrade to "3.1.7".

georgeconstantinou: @mamchenkov `~3.1` and `^3.1` does not mean it will upgrade up to 4.0.0?

mamchenkov: @georgeconstantinou nope ... have a look at this: 

https://getcomposer.org/doc/articles/versions.md#next-significant-release-operators

georgeconstantinou: I see, less than `4.0.0`

georgeconstantinou:
@mamchenkov please review

PR #25, opened 2016-01-28
Review and upgrade Phakefile (Task #796)

georgeconstantinou:
- Cache clearing
- Database migrations


.env.example
```diff
@@ -7,6 +7,7 @@ TEMPLATE_SRC=etc/nginx.conf.template
 TEMPLATE_DST=etc/nginx.conf
 
 DEBUG=1
+SYSTEM_COMMAND_COMPOSER="composer -n"
```
mamchenkov: @georgeconstantinou you shouldn't need it, as this is already defined in phake-builder.  See this:

https://github.com/QoboLtd/phake-builder/blob/master/lib/Composer.php#L11

georgeconstantinou: @mamchenkov I do not want the `--no-dev` as it skips composer dev packages like DebugKit.

The problem resides that by default we set `DEBUG=1`, which in turn enables debug mode in CakePHP, which in turn it tries to load the DebugKit plugin which does not exists because of the `--no-dev` flag.

mamchenkov: @georgeconstantinou I think we should disable the debug by default then.  The configuration for the template should be production-ready.  If the developer using the template needs debug or dev tools, he can change that in his local .env file.

Phakefile
```diff
@@ -10,6 +10,7 @@ group('app', function() {
 	});
 	task('install', ':git:pull', ':git:checkout');
 	task('install', ':dotenv:create', ':dotenv:reload', ':file:process');
+	task('install', ':cakephp:install:run');
```
mamchenkov: @georgeconstantinou I don't think so many nesting levels are needed.  Have a look at the project-template-wordpress, for example.  Having ```cakephp:install``` and ```cakephp:upgrade``` should be sufficient, and it will follow the logic of the ```app``` and other templates.

.env.example
georgeconstantinou: @mamchenkov sure, I will fix it.

Phakefile
georgeconstantinou: @mamchenkov :ok:

flangofas:
:+1:

flangofas:
@georgeconstantinou @mamchenkov i have added clear cache for the persistent folder as well

Phakefile
```diff
@@ -31,5 +33,68 @@ group('app', function() {
...
+		$command = getenv('CAKE_CONSOLE') . ' clear_cache all';
+		doShellCommand($command);
+	});
+
+	/**
+	 * 'Grouped CakePHP app update related tasks
+	 */
+	desc('Runs CakePHP app update related tasks');
+	task(
+		'update',
```
mamchenkov: @georgeconstantinou I think you also need ':builder:init' dependency.  These tasks will work if they are called from others which already have this dependency (like 'app:install'), but will probably fail if called directly from phake (`./vendor/bin/phake cakephp:install`).

.env.example
```diff
@@ -21,3 +21,6 @@ DB_DUMP_PATH=etc/mysql.sql
 
 GIT_REMOTE=origin
 GIT_BRANCH=
+
+CAKE_BIN_DIR=bin
+CAKE_CONSOLE="${CAKE_BIN_DIR}/cake"
```
mamchenkov: @georgeconstantinou I don't think there is a need for `CAKE_BIN_DIR`.  We already have `bin/` folder from the upstream project-template and it's not going anywhere.  `CAKE_CONSOLE` is sufficient.  It probably won't change ever either, but fine. :)

src/Shell/ClearCacheShell.php
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    /**
+     * Clear cache for models
+     *
+     * @return bool
+     */
+    public function models()
+    {
+        $conMan = ConnectionManager::get('default');
```
mamchenkov: @georgeconstantinou I think we should just clear [all configured Cache](https://github.com/QoboLtd/project-template-cakephp/blob/master/config/app.php#L101) folders.  Irrelevant of the existence of tables, etc.  Plugins can configure additional cache locations for unpredictable needs - generated files, and things like that.  There is no reliable way for us to know about all of them, except through the Cache configurations.

src/Shell/QoboPluginShell.php
```diff
@@ -0,0 +1,39 @@
...
+namespace App\Shell;
+
+use Cake\Console\ConsoleOptionParser;
+use Cake\Shell\PluginShell;
+
+/**
+ * Custom Plugin Shell class that adds extended
+ * functionality to Cake's core Plugin Shell.
+ */
+class QoboPluginShell extends PluginShell
```
mamchenkov: @georgeconstantinou you can name it PluginShell, so that it has the same name as the core plugin. I think it will make it easier for people to see that "app plugin" extends "core plugin".

src/Shell/Task/MigrateTask.php
```diff
@@ -0,0 +1,64 @@
...
+     * Method that runs migrations of all loaded plugins.
+     * @param  array $plugins loaded plugins
+     * @return void
+     */
+    protected function _migratePlugins(array $plugins)
+    {
+        foreach ($plugins as $plugin) {
+            $this->out('Migrating plugin [' . $plugin . ']:');
+            $this->dispatchShell([
+               'command' => ['migrations', 'migrate', '--plugin', $plugin]
```
mamchenkov: @georgeconstantinou CakePHP 3 migrations implement 4 commands (see [here](http://book.cakephp.org/3.0/en/migrations.html)):

* migrate
* rollback
* status
* mark_migrated

If these are too much to implement right now, we should at least provide a comment/TODO somewhere.

georgeconstantinou:
@mamchenkov please review

mamchenkov:
@georgeconstantinou @flangofas good work guys! Let's merge this in and move on.  If we need to improve it, we can do it a later stage.

mamchenkov:
@flangofas I guess you can proceed with your part now for the testing stuff.

flangofas:
:+1:

PR #28, opened 2016-02-02
	Add Test Database in app config (task #797)

flangofas:


.env.example
```diff
@@ -16,6 +16,7 @@ DB_HOST=localhost
 DB_USER=root
 DB_PASS=
 DB_NAME=
+DB_NAME_TEST=
```
mamchenkov: @flangofas no need for this.  Just use DB_NAME with the '_test' suffix.

flangofas: @mamchenkov i already used it twice in different places. I believe it is better to assign it to a configuration variable instead of concatenating it here and there. No?

mamchenkov: @flangofas to make it consistent, you'd need to provide all the other DB variables for the test database as well (username, password, host, etc).  Every configuration variable increases the complexity of the initial setup and things going wrong.  I think for us it's safe to assume that we'll be creating the test database based on the naming convention of the default database, using the same credentials, etc.

mamchenkov: @flangofas if you don't want to redefine the same thing in different places, you can read the configuration from the same source.  Since you do have to have it in `app/config.php`, you can read that from the `Phakefile`.

flangofas: @mamchenkov I have sent a fix. Just saw this. I have to check how this is done with Phakefile.

mamchenkov: @flangofas you still have it in `.env.example`.

flangofas: @mamchenkov removed from .env :smile:

PR #29, opened 2016-02-03
Integrate user authentication mechanisms (task #804)

georgeconstantinou:


mamchenkov:
@georgeconstantinou can you please adjust the unit tests as well.

Phakefile
```diff
@@ -88,6 +90,15 @@ group('cakephp', function() {
 		doShellCommand($command);
 	});
 
+	desc('Create \'qobo\' user');
+	task('qobo_user', ':builder:init', function() {
+		printSeparator();
+		printInfo('Creating \'qobo\' user');
+
+		$command = getenv('CAKE_CONSOLE') . ' users addUser --username=qobo --password=qobo --email=webdev@qobocloud.com';
```
flangofas: @georgeconstantinou Is this user super admin?

mamchenkov: @georgeconstantinou also, the values of the username, password, and email should come from the `.env` file, similar to how we have in wordpress template.

georgeconstantinou:
@mamchenkov please review

Phakefile
```diff
@@ -88,6 +90,18 @@ group('cakephp', function() {
 		doShellCommand($command);
 	});
 
+	desc('Create \'qobo\' user');
```
mamchenkov: @georgeconstantinou maybe 'Create dev user' or 'Create default user', to avoid quotes altogether. Especially since now 'qobo' user can be changed to anything via the .env.

georgeconstantinou: @mamchenkov done

PR #30, opened 2016-02-04
Update readme

flangofas:


mamchenkov:
@flangofas why are you trying to scare people away ? :)  :ghost:

flangofas:
@mamchenkov since we are adding stuff in and i thought it is good idea to have this in the readme file. In case they clone the master branch :smile:

mamchenkov:
@flangofas I don't think we need it ... those who want stable, should work with releases.  And having this in the README means that it will be cloned into all those other projects as well.  Scaring all those other people who forgot to update the README :)

PR #32, opened 2016-02-04
Requiring qobo CakePHP groups plugin (task #805)

georgeconstantinou:
:information_source: Pushed also groups plugin history just for reference. The plugin already moved to its own repo and is being installed through composer.


composer.json
```diff
@@ -5,34 +5,37 @@
...
+            "App\\": "src",
+            "Groups\\": "./plugins/Groups/src"
         }
     },
     "autoload-dev": {
         "psr-4": {
             "App\\Test\\": "tests",
-            "Cake\\Test\\": "./vendor/cakephp/cakephp/tests"
+            "Cake\\Test\\": "./vendor/cakephp/cakephp/tests",
+            "Groups\\Test\\": "./plugins/Groups/tests"
```
mamchenkov: @georgeconstantinou do you want to always run the plugin tests as part of the template?  Or should those tests be separate?

georgeconstantinou: @mamchenkov no need. I will remove it from here.

PR #40, opened 2016-03-11
require debugkit plugin

georgeconstantinou:


flangofas:
@georgeconstantinou Why you added debugkit there?

georgeconstantinou:
@flangofas to avoid deployment issues when debug flag is on, see today's conversation https://qobo.hipchat.com/chat/room/2429866 from time 11:15 and onwards.

flangofas:
:+1:

PR #46, opened 2016-03-23
change templates layout to match sb admin styling (task #1577)

georgeconstantinou:
Affected templates:
* register
* profile
* change password


flangofas:
:+1:

PR #56, opened 2016-04-13
prevent site rendering in iframe (task #1735)

georgeconstantinou:


src/Controller/AppController.php
```diff
@@ -54,6 +54,11 @@ public function initialize()
      */
     public function beforeFilter(Event $event)
     {
+        /*
+        Prevent page rendering in iframe
+         */
+        $this->response->header('X-Frame-Options', 'DENY');
```
mamchenkov: @georgeconstantinou for our case, it's better to have a .env.example variable set to DENY, rather then hardcode it, as there might be cases where the application will need to be inside an iframe.

georgeconstantinou:
@mamchenkov please review again.

flangofas:
:+1:

leonidsl:
:+1:

PR #57, opened 2016-04-13
Add session inactive timeout configuration (task #1734)

flangofas:


leonidsl:
:+1:

config/app.php
```diff
@@ -21,6 +21,7 @@
 $sessionCookieSecure = (bool)env('APP_SESSION_SECURE_COOKIE');
 $sessionCookieSecure = $https ?: $sessionCookieSecure;
 $cookieHttpOnly = (bool)env('APP_SESSION_COOKIE_HTTP_ONLY');
+$sessionTimeout = env('APP_SESSION_TIMEOUT');
```
mamchenkov: @flangofas `$sessionTimeout = (int)env('APP_SESSION_TIMEOUT');` seems appropriate ;)

flangofas: @mamchenkov Empty strings will be 0 then.

mamchenkov: @flangofas yup, and 0 disables session timeout, if I remember correctly.

flangofas: @mamchenkov disable session timeout or default to php.ini?

mamchenkov: @flangofas I think it 0 means disable the timeout ... who handles the session - php or cakephp - is a different setting.

flangofas:
@mamchenkov please review.

PR #76, opened 2016-06-01
Update project's bootstrap (task #2070)

georgeconstantinou:


mamchenkov:
@georgeconstantinou I was thinking about this the other day actually.  This is getting messy and complicated. :)  I think we can just have a list of plugins, and then we can find if the plugin has routes and bootstrap files, and if it has, then load it ... something like:

```php
$plugins = array(
'DebugKit',
// ...
'CakeDC.Users',
);
foreach ($plugins as $plugin) {
  $hasBootstrap = pluginHasBootstrap($plugin);
  $hasRoutes = pluginHasRoutes($plugin);
  Plugin::load($plugin, ['bootstrap' => $hasBootstrap, 'routes' => $hasRoutes]);
}
```

What do you think?

PR #85, opened 2016-08-08
Merge required commits

georgeconstantinou:


src/Controller/AppController.php
```diff
@@ -40,10 +45,15 @@ class AppController extends Controller
     public function initialize()
     {
         parent::initialize();
+        $this->loadComponent('RequestHandler');
         $this->loadComponent('Flash');
-        $this->loadComponent('Csrf');
+        // Temporarily disabled. @todo Find out why this is returning 403 on API calls.
+        // $this->loadComponent('Csrf');
```
mamchenkov: @georgeconstantinou because API requests come with the same token as was given for the page, but the tokens are one-use only.  You can disable the CSRF tokens for API calls for now, or for calls to specific form methods.

src/Controller/AppController.php
```diff
@@ -55,6 +64,19 @@ public function initialize()
...
     public function beforeFilter(Event $event)
     {
+        /*
+        if user not logged in, redirect him to login page
+         */
+        try {
+            $this->_checkAccess($event);
+        } catch (ForbiddenException $e) {
+            if (empty($this->Auth->user())) {
+                $this->redirect('/login');
```
mamchenkov: @georgeconstantinou above you have defined `loginRedirect` as `/`. Here you are using `/login`.  Which one is it and why do we need it twice? :)

georgeconstantinou: `loginRedirect` is for after user authentication redirect. The `/login` part is for pre-authenticated sessions.

PR #97, opened 2016-09-13
Google analytics snippet (task #2449)

flangofas:


mamchenkov:
@flangofas 

1. `analytics` should be `google_analytics`
2. Why do you need to put the `.env` value into configuration, when you can read it directly from the `.env`?

flangofas:
@mamchenkov I thought you were already happy with this PR as you have already reviewed it in another repository.

mamchenkov:
@flangofas my attention is very scattered today - I don't notice all the issues at once. Sorry about that.

I think the other PR also mentioned have a more specific `google_analytics` name, rather than a generic `analytics`, which can overlap with several different tools.

PR #106, opened 2016-09-21
Swagger Integration (task #2246)

georgeconstantinou:


src/Controller/Api/AppController.php
```diff
@@ -3,6 +3,16 @@
 
 use CsvMigrations\Controller\Api\AppController as BaseController;
 
+/**
+    @SWG\Swagger(
+        @SWG\Info(
+            title="PROJECT_NAME API Documentation",
```
mamchenkov: @georgeconstantinou let's just call it "API Documentation" for now, until we can dynamically throw the PROJECT_NAME in there.

src/Controller/Api/AppController.php
```diff
@@ -3,6 +3,16 @@
...
+            description="The PROJECT_NAME API allows interaction with PROJECT_NAME PROJECT_TYPE from your own application.",
```
mamchenkov: @georgeconstantinou here as well, we can just say "Interactive API documentation powered by Swagger.io" :)

PR #123, opened 2016-10-18
Upgraded cakephp-qobo-admin-panel plugin to v2.0.0

georgeconstantinou:


georgeconstantinou:
:information_source: CI tests were run locally as TravisCI seems to be down.

PR #132, opened 2016-10-26
Added plugins' tests to PHPUnit (task #2885)

mamchenkov:
Added plugins that can run standalone tests to the PHPUnit to
improve integration testing.


mamchenkov:
The tests are failing because of the underlying plugins.  The fixes to the plugins were applied and merged.  Once the new versions are tagged, the TravisCI run should be repeated to take care of the failures.

PR #142, opened 2016-11-16
Active Directory integration (task #2542)

georgeconstantinou:


.env.example
```diff
@@ -58,6 +58,13 @@ EMAIL_DOMAIN=localhost
 EMAIL_FROM_ADDRESS=root@localhost
 EMAIL_FROM_NAME="System Administrator"
 
+# LDAP configuration
+LDAP_AUTHENTICATION=0
```
mamchenkov: @georgeconstantinou please change 'authentication' to 'enabled'.  It'll be clearer, as 'authentication' can imply algorithms, modes, etc.

georgeconstantinou: done 755e51d

PR #144, opened 2016-11-18
Add LDAP search support (task #2542)

georgeconstantinou:


.env.example
```diff
@@ -65,6 +65,8 @@ LDAP_VERSION=3
 LDAP_HOST=
 LDAP_PORT=389
 LDAP_BASE_DN=
+LDAP_FILTER=
+LDAP_ATTRIBUTES=
```
mamchenkov: @georgeconstantinou please set LDAP_FILTER to the samAccountName and the LDAP_ATTRIBUTES to givenname, sn, and mail.

georgeconstantinou: @mamchenkov 004779b

PR #148, opened 2016-11-23
Import shell (task #2503)

mamchenkov:
First draft of the Import Shell.  Currently, it only knows how to generate CSV template files and their documentation.  Run like this:

```
mkdir /tmp/import
cake import create_templates /tmp/import
```


mamchenkov:
:+1:

PR #177, opened 2016-12-21
AdminLTE user panel overwrite (task #3151)

georgeconstantinou:


src/Template/Element/aside/user-panel.ctp
```diff
@@ -0,0 +1,15 @@
+<?php
+$displayName = $user['first_name'] . ' ' . $user['last_name'];
+if (empty(trim($displayName))) {
+    $displayName = $user['username'];
```
mamchenkov: @georgeconstantinou I think this is not necessary anymore.  We have the `$user['name']` coming from the User entity.

PR #225, opened 2017-03-16
Changed a way to call roles and capabilities method hasAccess from application

stepanov:


src/Event/Menu/MenuListener.php
```diff
@@ -108,6 +111,7 @@ protected function _getDashboardLinks(array $user)
      * @param  \Cake\Event\Event $event Event object
      * @param  array             $menu  Menu
      * @param  array             $user  User
+use RolesCapabilities\CapabilityTrait;
```
anvyst: @stepanov , looks like a paste in the middle of comment block. Can you have a look at it?

stepanov: @anvyst  removed. was added by mistake

PR #227, opened 2017-03-16
Update plugin migration to run it separately

stepanov:
Currently migration is run for all loaded plugins in one time. As result migrations files with the same class names cannot be used.

### DIFF
--- a/Phakefile.php
+++ b/Phakefile.php
@@ -188,9 +188,15 @@ group('cakephp', function () {
         doShellCommand($command, $secureStrings);
 
         // Run plugin migrations
-        printInfo("Testing plugin migrations");
-        $command = getenv('CAKE_CONSOLE') . ' plugin migrations migrate --connection=test';
-        doShellCommand($command);
+        printInfo("Testing ;plugin migrations");
+        $result = doShellCommand(getenv('CAKE_CONSOLE') . ' plugin loaded');
+        $result = preg_replace('/^.*\-\n/s', '', $result);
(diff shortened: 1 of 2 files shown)


Phakefile.php
```diff
@@ -187,11 +187,17 @@ function getProjectVersion($app = null)
         $secureStrings = ['DB_PASS', 'DB_ADMIN_PASS'];
         doShellCommand($command, $secureStrings);
 
-        // Run plugin migrations
-        printInfo("Testing plugin migrations");
-        $command = getenv('CAKE_CONSOLE') . ' plugin migrations migrate --connection=test';
-        doShellCommand($command);
-
+        // Run plugin migrations separately for each plugin
+        printInfo("Testing ;plugin migrations");
```
mamchenkov: @stepanov there's a typo (semicolon)

Phakefile.php
```diff
@@ -187,11 +187,17 @@ function getProjectVersion($app = null)
...
+        $command = getenv('CAKE_CONSOLE') . ' plugin loaded';
+        $result = doShellCommand($command);
+        $result = preg_replace('/^.*\-\n/s', '', $result);
```
mamchenkov: @stepanov the way to avoid the printout is to extend the shell in the app, with the following method:

```
/**
 * No welcome message
 *
 * @return void
 */
protected function _welcome()
{
}
```

See, for example, vendor/cakephp/cakephp/src/Shell/Task/ExtractTask.php

PR #246, opened 2017-05-04
Added default menu item icon in case of table does not have it

stepanov:


config/menu.php
```diff
@@ -3,6 +3,8 @@
 return [
     'Menu' => [
         // call getMenu method only once
-        'allControllers' => false
+        'allControllers' => false,
+        // Default icon for menu item
+        'default_menu_item_icon' => 'bars',
```
mamchenkov: @stepanov `cube` is out default icon for modules which don't specify the icon.

stepanov: @mamchenkov changed

PR #248, opened 2017-05-04
Added manage language into system menu as per task #806

stepanov:


config/Modules/Common/config/menus.json
```diff
@@ -0,0 +1,60 @@
...
+        },
+        {
+            "label": "Information",
+            "desc": "System information screen",
+            "url": "/system/info/",
+            "icon": "info-circle bg-light-blue",
+            "order": 70
+        },
+        {
+            "label": "Settings",
```
mamchenkov: @georgeconstantinou @anvyst @stepanov do you guys remember if there is a reason for this menu item?  We don't have this functionality yet, but the menu item shows up everywhere.

Can we remove it until we have the functionality, or does it affect something somewhere?

georgeconstantinou: I think it was used just as a placeholder.

stepanov: @mamchenkov removed

PR #249, opened 2017-05-05
Menu items access check (task #3757)

georgeconstantinou:
Bugfix for children items not being checked, if parent item is an external link.

### DIFF
--- a/src/Event/Menu/MenuListener.php
+++ b/src/Event/Menu/MenuListener.php
@@ -179,36 +179,34 @@ class MenuListener implements EventListenerInterface
     protected function _checkItemAccess(array $items, array $user)
     {
         $fullBaseUrl = Router::fullBaseUrl();
+
         foreach ($items as $k => &$item) {
             $url = $item['url'];
-            if (is_string($url)) {
-                // skip access check on external links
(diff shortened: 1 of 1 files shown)


src/Event/Menu/MenuListener.php
```diff
@@ -179,44 +179,42 @@ protected function _checkItemsAccess(Event $event, array $menu, array $user)
...
-            if (is_string($url)) {
-                // skip access check on external links
-                if (preg_match('/http/i', $url) && (0 !== strpos($url, $fullBaseUrl))) {
-                    continue;
-                }
-                // strip out full base URL if is part of menu item's URL
+            $internal = true;
+
+            // skip access check on external links
+            if (is_string($url) && preg_match('/http/i', $url) && (0 !== strpos($url, $fullBaseUrl))) {
```
mamchenkov: @georgeconstantinou the method is already long enough.  This can be moved out into a method of its own :)

georgeconstantinou: done https://github.com/QoboLtd/project-template-cakephp/pull/249/commits/12b9f653b8e01592993e8f6996d326eaf4098350

PR #253, opened 2017-05-10
Add 'listing' of dblists to everyone (task #3798)

anvyst:
Dblists might block access to view the page, in case the user doesn't have the capability to 'list' dblists or dblist items.


mamchenkov:
@anvyst I think this should be done as the shell script rather than a migration.  We already have shell scripts that handle defaults roles, menus, etc, so we should add it there and run it during the update/install process.  Otherwise, if somebody unsets the permission, the breakage will ensure.

PR #255, opened 2017-05-11
Moving migration dblist scheck to shell (task #3798)

anvyst:
To have more control over permissions checks, Migration script
is moved to the shell so we can call it whenever we need.

### DIFF
--- a/Phakefile.php
+++ b/Phakefile.php
@@ -275,6 +275,7 @@ group('cakephp', function () {
             'role import',
             'capability assign',
             'menu import',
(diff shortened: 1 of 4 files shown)


Phakefile.php
```diff
@@ -275,6 +275,7 @@ function getProjectVersion($app = null)
             'role import',
             'capability assign',
             'menu import',
+            'check_dblist_permissions',
```
stepanov: Please change it to add_dblist_permissions

PR #264, opened 2017-05-18
Added report grid event listener (task #3805)

stepanov:


src/Event/View/ReportGridViewListener.php
```diff
@@ -0,0 +1,44 @@
...
+
+use Cake\Event\Event;
+use Cake\Event\EventListenerInterface;
+use Cake\Log\LogTrait;
+use Cake\Network\Exception\ForbiddenException;
+use CsvMigrations\FieldHandlers\FieldHandlerFactory;
+
+class ReportGridViewListener implements EventListenerInterface
+{
+    use LogTrait;
```
mamchenkov: @stepanov logging is not used anymore, so maybe the trait should be removed as well.

PR #278, opened 2017-06-27
Added MagicDefaultValueListener (task #3567)

mamchenkov:
This even listener, combined with the latest
CsvMigrations plugin, allows for dynamic values
as field defaults (which is magical).

For example, instead of the static defaults like

```
[field]
default=foobar
```

One now can set dynamic values like:

```
[field]
default=%CURRENT_DATE%
```

Supported dynamic defaults are:

* CURRENT_DATE
* CURRENT_TIME
* CURRENT_DATETIME
* CURRENT_USER_ID

More can be easily supported by adding protected methods
to the MagicDefaultValueListener, matching the naming
convention:

1. Convert default to lower case.
2. Convert from underscores to CamelCase.
3. Prefix with 'get'.
4. Suffix with 'Value'.

For example, `CURRENT_USER_ID` becomes `getCurrentUserIdValue()`.

### DIFF
--- a/config/bootstrap.php
+++ b/config/bootstrap.php
@@ -39,6 +39,7 @@ if (!extension_loaded('intl')) {
 }
 
 use App\Event\Component\UserIdentifyListener;
+use App\Event\MagicDefaultValueListener;
 use App\Event\Menu\MenuListener;
 use App\Event\Model\SearchableFieldsListener;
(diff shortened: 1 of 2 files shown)


mamchenkov:
This PR won't change any behavior until https://github.com/QoboLtd/cakephp-csv-migrations/pull/421 is merged and tagged.

PR #309, opened 2017-08-09
New way to render buttons and action buttons

stepanov:


src/Event/View/BaseMenuListener.php
```diff
@@ -18,14 +20,36 @@
      * @param  \Cake\Event\Event $event Event object
      * @param  array             $menu  Menu
      * @param  array             $user  User
+     * @param  string            $type  menu type
      * @return void
      */
-    public function beforeRenderFlatMenu(Event $event, array $menu, array $user)
+    public function beforeRenderFlatMenu(Event $event, array $menu, array $user, $type = 'buttons')
```
mamchenkov: @stepanov it's better to use constants, rather than hardcoded strings like 'buttons'.

src/Event/View/BaseMenuListener.php
```diff
@@ -18,14 +20,36 @@
...
+                continue;
+            }
+
+            $menuItem = MenuItemFactory::createMenuItem($item);
+            $menuBuilder->addMenuItem($menuItem);
+        }
+
+        if ($type == 'actions') {
+            $renderClass = 'Menu\\MenuBuilder\\MenuActionsRender';
+        } else {
```
mamchenkov: @stepanov you can get rid of the `else` block by setting `$renderClass` to the default value of buttons, and then overwriting in the `if` block.  Or just generate the renderer class name from the parameter, and checking that the class exists.  This way you won't have to update the code when support for more than two types of rendering is necessary.

src/Event/View/BaseMenuListener.php
```diff
@@ -18,14 +20,36 @@
...
+            $renderClass = 'Menu\\MenuBuilder\\MenuButtonsRender';
+        }
+
+        $render = new $renderClass($menuBuilder, $event->subject());
+        $event->result = $render->render();
+
+        return;
+
```
mamchenkov: @stepanov you have code leftover after the unconditional `return` statement.  The execution will never get there, so this code will never run.  Please remove.

mamchenkov:
Looks fine to me, but better if @georgeconstantinou and @anvyst review it as well before merging.

georgeconstantinou:
Approved by me, @anvyst can have a look too

PR #311, opened 2017-08-11
Fix redirect after profile edit - task #4335

stepanov:


src/Controller/UsersController.php
```diff
@@ -117,4 +118,39 @@ public function uploadImage($id)
...
+     * @return void
+     */
+    public function editProfile()
+    {
+        $this->autoRender = false;
+        $this->request->allowMethod(['patch', 'post', 'put']);
+
+        $user_id = $this->Auth->user('id');
+        if (empty($user_id)) {
+            throw new UnauthorizedException('You have tpo login to complete this action!');
```
georgeconstantinou: @stepanov there is a typo in the exception message: `tpo`

georgeconstantinou: @stepanov there is a typo in the exception message: `tpo`

PR #340, opened 2017-08-30
Nightwatch to do end-2-end UI tests (task #4208)

anvyst:


mamchenkov:
@anvyst setting to [WIP] for now ... we need to think and play around with it more.  It'll probably end up in project-template, so that we can inherit it into the project-template-wordpress as well.

PR #348, opened 2017-09-08
Added new field - extras to the table users to store additional user info

stepanov:


config/Migrations/20170908100742_AddFieldExtrasToUsers.php
```diff
@@ -0,0 +1,23 @@
...
+     * More information on this method is available here:
+     * http://docs.phinx.org/en/latest/migrations.html#the-change-method
+     * @return void
+     */
+    public function change()
+    {
+        $table = $this->table('users');
+        $table->addColumn('extras', 'text', [
+            'default' => null,
+            'limit' => 4294967295,
```
mamchenkov: @stepanov it's better to use a constant here ... have a look at `config/Migrations/20160324102121_CreateLogAudit.php` and `config/Migrations/20160329150805_AddOriginalToLogAudit.php`.

PR #370, opened 2017-11-06
Added possibility to refresh session after profile update

stepanov:


config/users.php
```diff
@@ -6,6 +6,10 @@
     // disable remember-me functionality because currently it affects google authenticator functionality:
     // https://github.com/CakeDC/users/issues/488
     'Users.RememberMe.active' => false,
+    'Users.Email.required' => true,
+    'Users.Registration.active' => true,
+    'Users.Email.validate' => true,
+    'Users.Tos.required' => false,
```
mamchenkov: @stepanov email should not be required, registration should not be active, and the email shouldn't not be validated by default.

PR #375, opened 2017-11-13
Initial Feature Management implementation (task #4839)

georgeconstantinou:
This PR includes the initial Feature Management implementation.

**Feature Types**
Feature types can be created under `src/Feature/Type` directory. The class name must be suffixed with "Feature". Additionally, it has to extend `App\Feature\Type\BaseFeature` class or implement `App\Feature\FeatureInterface` interface.
```php
<?php
namespace App\Feature\Type;

use App\Feature\Type\BaseFeature;

class AwesomeFeature extends BaseFeature
{
    public function enable()
    {
        // code that enables Awesome feature 
    }

    public function disable()
    {
        // code that disables Awesome feature 
    }
}
```

**Features configuration**
A feature configuration must include the feature name (string) and active flag (boolean). Configuration must be defined in `config/features.php`. You can also pass additional parameters, which can be accessed within your Feature class, using the `$config->get('my-parameter')` method. Below is a sample configuration code:
```php
<?php
use App\Feature\Feature;

// Features configuration
return [
    'Features' => [
        'Awesome' => ['active' => false],
    ],
];
```

**Managing features**
Below is a code example of how to manage features:
```php
use App\Feature\Factory;

/**
 * Below, we run the Factory init method, which will loop through all feature items
 * in the configuration, and based on their active status, it will run the relevant
 * enable or disable method on their respective Feature classes.
 */
Factory::init();

/**
 * Additionally, we can retrieve a specific feature and check its active status
 * to build our application logic.
 */
$feature = Factory::get('Awesome');
if (!$feature->isActive()) {
    throw new \RuntimeException('This is not awesome.');
}
```


src/Feature/Collection.php
```diff
@@ -27,6 +35,32 @@ public function __construct(array $data)
     }
 
     /**
+     * Normalize features configuration data.
+     *
```
mamchenkov: @georgeconstantinou can you please provider more information on what is actually being done here ... 'normalize' is a very vague term. :)

georgeconstantinou: @mamchenkov https://github.com/QoboLtd/project-template-cakephp/pull/375/commits/de63b8234379cb5e6cebb1a3f6b8bb0aa992b411

src/Feature/Config.php
```diff
@@ -0,0 +1,63 @@
...
+
+    /**
+     * Constructor method.
+     *
+     * @param array $config Feature config
+     */
+    public function __construct(array $config)
+    {
+        $name = Hash::get($config, 'name');
+        if (!is_string($name)) {
```
mamchenkov: @georgeconstantinou what if the name is an empty string? :)

src/Feature/FeatureInterface.php
```diff
@@ -0,0 +1,26 @@
+<?php
+namespace App\Feature;
+
+interface FeatureInterface
+{
+    /**
```
mamchenkov: @georgeconstantinou I think the constructor should be part of the interface as well

config/features.php
```diff
@@ -0,0 +1,7 @@
+<?php
+// Features configuration
+return [
+    'Features' => [
+        // ['name' => 'Batch', 'active' => true],
```
mamchenkov: @georgeconstantinou can you have a correct example commented here or remove the comment at all.  Thanks.

src/Controller/AppController.php
```diff
@@ -127,13 +136,33 @@ public function beforeFilter(Event $event)
...
         // Load AdminLTE theme
         $this->loadAdminLTE();
     }
 
     /**
+     * Enable/disable features.
+     *
+     * @return void
+     */
+    protected function manageFeatures()
```
mamchenkov: @georgeconstantinou I don't think we need this now. :)

PR #382, opened 2017-11-21
Features improvements (task #4839)

georgeconstantinou:
- Added support for feature classes in sub-directories.
- Moved Factory initialization to bootstrap file.

### DIFF
--- a/src/Controller/Api/AppController.php
+++ b/src/Controller/Api/AppController.php
@@ -111,7 +111,7 @@ class AppController extends Controller
         ]);
 
         // prevent access on disabled module
-        $feature = FeatureFactory::get($this->name);
+        $feature = FeatureFactory::get('Module/' . $this->name);
         if (!$feature->isActive()) {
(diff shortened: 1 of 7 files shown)


src/Controller/Api/AppController.php
```diff
@@ -110,11 +110,8 @@ public function initialize()
             'currentRequest' => $this->request->params
         ]);
 
-        // Feature Factory initialization
-        FeatureFactory::init();
-
         // prevent access on disabled module
-        $feature = FeatureFactory::get($this->name);
+        $feature = FeatureFactory::get('Module/' . $this->name);
```
mamchenkov: @georgeconstantinou can we use DS instead of slash? Just in case somebody will ever try to use it on Windows :)

georgeconstantinou: @mamchenkov done in PR https://github.com/QoboLtd/project-template-cakephp/pull/383

PR #395, opened 2017-12-01
Adding tiny eye-candy on error 4xx and 5xx for user UI (task #4982)

anvyst:
An attempt to prettify existing error400 and error500 view files to make them use default template's layout.

### DIFF
--- a/src/Template/Error/error400.ctp
+++ b/src/Template/Error/error400.ctp
@@ -2,7 +2,7 @@
 use Cake\Core\Configure;
 use Cake\Routing\Router;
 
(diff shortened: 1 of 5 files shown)


src/Template/Error/error400.ctp
```diff
@@ -27,16 +27,20 @@ if (Configure::read('debug')):
     endif;
 
     $this->end();
-endif;
+} else {
+    $this->layout = 'error';
```
mamchenkov: @anvyst this can go above the `if` condition, just to avoid the `else` part.

PR #397, opened 2017-12-04
Fix some phpcs warnings

georgeconstantinou:


src/Controller/UsersController.php
```diff
@@ -127,19 +127,21 @@ public function uploadImage($id)
      * separate user record update by admin and editing profile
      * by logged in user
      *
-     * @return void
+     * @return \Cake\Network\Response
+     * @throws UserNotFoundException When user not found.
```
mamchenkov: @georgeconstantinou please use full path to class, like ```\Foo\Bar\Exception```.

PR #399, opened 2017-12-04
Modifying ExceptionRenderer for Logged users (task #4982)

anvyst:
We don't want to break user navigation for loggedin users.
In case we get 4xx/5xx error, we replace the content of the
page but keepting sidebars and nav-bar components.

It works only for disabled debug. Otherwise, the page
will be using native CakePHP error templates.

### DIFF
--- a/src/Error/AppExceptionRenderer.php
+++ b/src/Error/AppExceptionRenderer.php
@@ -36,7 +36,8 @@ class AppExceptionRenderer extends ExceptionRenderer
                     '_serialize' => ['message', 'code', 'url'],
                 ];
 
-                $res = $this->controller->request->session()->write('currentError', json_encode($data));
+                // adding generated error info into custom session variable for system/error page.
+                $this->controller->request->session()->write('currentError', json_encode($data));
 
                 return $this->controller->redirect('/system/error');
             }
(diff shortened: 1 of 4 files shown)


src/Error/AppExceptionRenderer.php
```diff
@@ -0,0 +1,48 @@
...
+     * @return mixed
+     */
+    public function render()
+    {
+        $isDebug = Configure::read('debug');
+        $this->controller->loadComponent('CakeDC/Users.UsersAuth');
+        $this->controller->Auth->config('authorize', false);
+        $currentUser = $this->controller->Auth->user();
+
+        if (!$isDebug) {
```
mamchenkov: @anvyst can this be simplified to avoid nested conditions, etc ... just return things as early as possible.

src/Template/System/error.ctp
```diff
@@ -0,0 +1,40 @@
...
+
+use Cake\Routing\Router;
+
+if (empty($user)) {
+    $this->layout = 'error';
+}
+
+$currentError = $this->request->session()->read('currentError');
+
+$url = $code = $message = '';
```
mamchenkov: @anvyst in case for some reason the 'currentError' is not present, can we still show some defaults.  For example, code '000', and message 'Unknown error'.

PR #414, opened 2017-12-15
bug fix on lookup action listener (task #5079)

christosapl:
when altering query was overwriting the query variable with empty value

### DIFF
--- a/src/Event/Controller/Api/LookupActionListener.php
+++ b/src/Event/Controller/Api/LookupActionListener.php
@@ -143,7 +143,7 @@ class LookupActionListener extends BaseActionListener
         ]);
 
         // recursive call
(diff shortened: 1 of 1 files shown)


src/Event/Controller/Api/LookupActionListener.php
```diff
@@ -143,7 +143,7 @@ protected function _getRelatedModuleValues(CsvField $csvField, ServerRequest $re
         ]);
 
         // recursive call
-        $query = $this->_alterQuery($table, $query, $request);
+        $this->_alterQuery($table, $query, $request);
```
mamchenkov: @christosapl I'm not quite sure that this fixes anything.

PR #428, opened 2017-12-27
Plugin route count (task #4839)

georgeconstantinou:
A plugin route might consist two parts, for example on default actions such as index. So to properly filter plugin links during menu rendering we check for a URL count of at least two.


mamchenkov:
@georgeconstantinou Merging, despite of the TravisCI failures, which are caused by the CakePHP 3.5.9, as per [this issue](https://github.com/cakephp/cakephp/issues/11577).

PR #433, opened 2018-01-12
App Scheduled Jobs module (task #5165)

anvyst:
Scheduled tasks are easier to maintain on the application level, as it solves certain issues with access control, and the usability for the end user.

Scheduled Jobs are controlled by iCalendar configuration (`rlanvin/php-rrule` for Recurrence rules). And the jobs are executed on each cronjob iteration.

**Note**: for now the execution of the cronjob is done every minute (which is the minimal Frequency unit of the recurrence).

Currently we support only Cake Shells to be added to scheduled jobs with minimal exclusions. Recurrence rule is added as string for the time being. Later will be replaced with field handler from `qobo/cakephp-csv-migrations` plugin


config/Modules/ScheduledJobs/config/config.ini
```diff
@@ -0,0 +1,6 @@
+[table]
+icon = cube
```
mamchenkov: @anvyst better icon? :)

src/Controller/ScheduledJobsController.php
```diff
@@ -0,0 +1,83 @@
...
+use CsvMigrations\Controller\AppController as BaseController;
+
+/**
+ * ScheduledJobs Controller
+ *
+ */
+class ScheduledJobsController extends BaseController
+{
+    /**
+     * {@inheritDoc}
```
mamchenkov: @anvyst can we have real docs instead of inheritdocs please? :)

src/Model/Table/ScheduledJobLogsTable.php
```diff
@@ -0,0 +1,119 @@
...
+    /**
+     * Log Scheduled Job script
+     *
+     * @param \Cake\Datasource\EntityInterface $entity of the scheduled job
+     * @param array $state of the script response
+     * @param \Cake\I18n\Time $stamp of currently executed cron iteration
+     *
+     * @return mixed $result containing bool or inserted Id
+     */
+    public function log(EntityInterface $entity, array $state, Time $stamp)
```
mamchenkov: @anvyst this probably clashes with the generic CakePHP logger ...

src/Shell/CronShell.php
```diff
@@ -0,0 +1,72 @@
...
+        foreach ($jobs as $entity) {
+            $rrule = $this->ScheduledJobs->getRRule($entity);
+            $shouldRun = $this->ScheduledJobs->timeToInvoke($now, $rrule);
+
+            if (!$shouldRun) {
+                continue;
+            }
+
+            $instance = $this->ScheduledJobs->getInstance($entity->job, 'Job');
+            $state = $instance->run($entity->options);
```
georgeconstantinou: @anvyst `getInstance()` will return `null` if the `job` does not exist. So we need to either make sure an instance of a job is always returned (maybe `DefaultJob` or something), or skip the `run()` and probably `logJob()` altogether.

PR #441, opened 2018-01-17
Cache recent activity results (task #5215)

georgeconstantinou:


src/Template/Element/aside-control-sidebar.ctp
```diff
@@ -1,18 +1,27 @@
...
-    ->find('all')
-    ->limit(10)
-    ->where(['LogAudit.meta LIKE' => '%"user":"' . $user['id'] . '"%'])
-    ->distinct(['LogAudit.primary_key'])
-    ->order(['timestamp' => 'DESC'])
-    ->all();
+
+$cacheKey = 'recent_activity_' . $user['id'];
+$history = Cache::read($cacheKey);
+if (!$history) {
```
mamchenkov: @georgeconstantinou I think it's better to check for null or false (whatever an empty/non-existing cache returns).  So that if we cache an empty array of the user history, it lasts for the cache timeout as well.  This should help for those users who have read-only access to the system, for example.

georgeconstantinou: @mamchenkov if the cache does not find the specified key, it will return false

mamchenkov: @georgeconstantinou yes.  So we should check for `false` then.  Currently, if the cache is found but contains an empty array for history - it will try to find the recent activity.  No?

georgeconstantinou: @mamchenkov $history holds the ResultSet instance, with or without results.

mamchenkov: @georgeconstantinou my bad :confused:

georgeconstantinou: @mamchenkov is still safer this way since we know that for non-existing caches the returned value will always be `false`.

PR #448, opened 2018-01-22
Abstract Module feature class (task #5228)

georgeconstantinou:


tests/TestCase/Feature/Type/ScheduledJobsFeatureTest.php
```diff
@@ -0,0 +1,31 @@
...
+use App\Feature\Factory;
+use Cake\Core\Configure;
+use Cake\TestSuite\TestCase;
+
+class ScheduledJobsFeatureTest extends TestCase
+{
+    public function testIsActive()
+    {
+        $config = Configure::read('Features.Module/ScheduledJobs');
+
```
mamchenkov: @georgeconstantinou what's with all the double spacing? :)

georgeconstantinou: that's from @anvyst

PR #454, opened 2018-02-05
Gravatar support for Users (task #5327)

georgeconstantinou:
Created avatar domain Service with support for `ImageSource` and `Gravatar`.

ImageSource loads the user's system-uploaded image or falls back to a default one if the user has not uploaded any image yet. You do 

Gravatar loads the user's avatar image by email, using the [gravatar](https://en.gravatar.com) API.

To enable Gravatar support you just need the following:

```php
<?php
// config/users.php

return [
    'Users.gravatar' => [
        'type' => 'App\Avatar\Type\Gravatar'
    ]
];
```

### DIFF
--- a/config/bootstrap.php
+++ b/config/bootstrap.php
@@ -284,4 +284,4 @@ FeatureFactory::init();
 /**
  * Register custom database type(s)
  */
-Type::map('base64image', 'App\Database\Type\EncodedImageType');
+Type::map('base64', 'App\Database\Type\EncodedFileType');
--- a/config/bootstrap.php
+++ b/config/bootstrap.php
@@ -59,6 +59,7 @@ use Cake\Core\App;
 use Cake\Core\Configure;
 use Cake\Core\Configure\Engine\PhpConfig;
 use Cake\Core\Plugin;
+use Cake\Database\Type;
 use Cake\Datasource\ConnectionManager;
 use Cake\Error\ErrorHandler;
 use Cake\Event\EventManager;
@@ -279,3 +280,8 @@ Configure::load('admin_lte', 'default');
 // Feature Factory initialization
 // IMPORTANT: this line should be placed at the end of the bootstrap file.
 FeatureFactory::init();
+
+/**
+ * Register custom database type(s)
+ */
+Type::map('base64image', 'App\Database\Type\EncodedImageType');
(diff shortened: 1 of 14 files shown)


config/bootstrap.php
```diff
@@ -279,3 +280,8 @@
 // Feature Factory initialization
 // IMPORTANT: this line should be placed at the end of the bootstrap file.
 FeatureFactory::init();
+
+/**
+ * Register custom database type(s)
+ */
+Type::map('base64image', 'App\Database\Type\EncodedImageType');
```
mamchenkov: @georgeconstantinou I think we can even make it more generic, as `base64`.  There might be other scenarios where we want to keep base64-encoded data in the database, not necessarily always an image.

config/users.php
```diff
@@ -16,4 +16,7 @@
         // 'CakeDC/Users.RememberMe',
         'Form',
     ],
+    'Users.gravatar' => [
+        'active' => false
+    ]
```
mamchenkov: @georgeconstantinou I think we should throw in the other options here as well, such as rating, size, and default option.

Here's a useful reference: https://en.gravatar.com/site/implement/images/

src/Model/Entity/User.php
```diff
@@ -30,19 +31,31 @@ protected function _getName()
...
+        if (Configure::read('Users.gravatar.active') && $this->get('email')) {
+            return $this->getGravatar($this->get('email'));
+        }
+
+        if ($this->get('image')) {
+            return $this->get('image');
         }
 
-        return $image;
+        return '/img/user-image-160x160.png';
```
mamchenkov: @georgeconstantinou before it made sense to be hardcoded.  But now that we have avatar related options, we can move this value into the configuration as well.

src/Model/Entity/User.php
```diff
@@ -30,19 +31,31 @@ protected function _getName()
...
+    }
+
+    /**
+     * Gravatar getter.
+     *
+     * @param string $email User email
+     * @return string
+     */
+    private function getGravatar($email)
```
mamchenkov: @georgeconstantinou I think this can be useful as a public function either in the Users Entity or standalone class.  (Fetching avatars by email address for those who are not system users, for example ... Leads/Accounts/etc).

Additionally, it would be nice to have a unit test for this functionality.  You can use the dev user email address in the unit test.

PR #459, opened 2018-02-10
Introducing new versioned API structure (task #2589)

anvyst:


config/routes.php
```diff
@@ -44,16 +44,9 @@
...
-     * handle json file extension on API calls
-     */
-    $routes->extensions(['json']);
-    $routes->resources('Users');
-    $routes->resources('LanguageTranslations');
-
-    $routes->fallbacks('DashedRoute');
-});
+if (!class_exists('ApiRoutes')) {
+    include('routes_api.php');
```
mamchenkov: @anvyst I think the class should be available via the common autoloader.  There are multiple uses for it - not just the routes loading, but also figuring out which versions of the API available.  (For things like Swagger documentation)

config/routes_api.php
```diff
@@ -0,0 +1,82 @@
...
+class ApiRoutes
+{
+    protected $_versions = [];
+
+    public function __construct() {
+        $this->_getVersions();
+        $this->_setRoutes();
+    }
+
+    protected function _getVersions() {
```
mamchenkov: @anvyst I think this is useful as a public function.

config/routes_api.php
```diff
@@ -0,0 +1,82 @@
...
+    {
+        return '/api/v' . $version;
+    }
+
+    protected function _getApiRoutePrefix($versions)
+    {
+        return 'api/v' . implode('/v', $versions);
+    }
+
+    protected function _setRoutes()
```
mamchenkov: @anvyst I think this should be a public function, called from the `config/routes.php`.

config/routes_api.php
```diff
@@ -0,0 +1,82 @@
...
+                Router::prefix($version['prefix'], ['path' => $version['path']], function ($routes) {
+                    $routes->extensions(['json']);
+                    $routes->fallbacks(DashedRoute::class);
+                });
+            }
+        });
+    }
+}
+
+new ApiRoutes();
```
mamchenkov: @anvyst I think this should be removed from the class file.

PR #469, opened 2018-02-13
Adding App check for migrated API controllers (task #5386)

anvyst:
Due to API versioning that use sub directories for Controllers, we should check that all of them were correctly placed in corresponding directories.

### DIFF
--- a/tests/TestCase/Controller/Api/ControllerApiTest.php
+++ b/tests/TestCase/Controller/Api/ControllerApiTest.php
@@ -9,19 +9,15 @@ class ControllerApiTest extends TestCase
 {
     public function testApiFilesPlacedCorrectly()
     {
-        $dir = App::path('Controller/Api')[0];
-        $dir = new Folder($dir);
-
-        $contents = $dir->read(true, true);
+        $path = App::path('Controller/Api')[0];
+        $dir = new Folder($path);
         $found = 0;
 
         // checking for scanned files
-        if (!empty($contents[1])) {
-            foreach ($contents[1] as $file) {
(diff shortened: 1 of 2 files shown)


tests/TestCase/Controller/Api/ControllerApiTest.php
```diff
@@ -0,0 +1,31 @@
...
+use Cake\TestSuite\TestCase;
+
+class ControllerApiTest extends TestCase
+{
+    public function testApiFilesPlacedCorrectly()
+    {
+        $dir = App::path('Controller/Api')[0];
+        $dir = new Folder($dir);
+
+        $contents = $dir->read(true, true);
```
mamchenkov: @anvyst nothing wrong with this, but I think this approach is slightly better - https://github.com/QoboLtd/project-template-cakephp/blob/master/src/Shell/UpgradeShell.php#L70

What do you think?

tests/TestCase/Controller/Api/ControllerApiTest.php
```diff
@@ -9,19 +9,15 @@ class ControllerApiTest extends TestCase
...
 
         // checking for scanned files
-        if (!empty($contents[1])) {
-            foreach ($contents[1] as $file) {
-                if (preg_match('/^(.*)Controller\.php$/', $file, $matches)) {
-                    if (count($matches) > 1) {
-                        $found++;
-                    }
+        foreach ($dir->find('^\w+Controller\.php$') as $file) {
+            if (preg_match('/^(.*)Controller\.php$/', $file, $matches)) {
```
georgeconstantinou: @anvyst this is not needed, as the `find()` method will return the relevant results, if found.

PR #472, opened 2018-02-14
Adding placeholders to directories.

fr3nch13:


mamchenkov:
@fr3nch13 thanks!

fr3nch13:
Sure. I like ‘.empty’ over ‘empty’ because it avoids a rm -f [dir]

-------------
Brian French
703-507-5383

mamchenkov:
@fr3nch13 indeed.  Having a hidden file is better.  However, I'll probably rename them to `.gitkeep` soon to make the purpose of the empty hidden file a little bit more obvious, especially for those people who are not well familiar with git.

fr3nch13:
True. The other pseudo standard was would be to rename them to .gitignore.  I like .gitkeep as it’s more intuitive. 

-------------
Brian French
703-507-5383

fr3nch13:
Off subject. I was looking at your other projects. Considering you’re dabbling in forex, have you guys been looking in to crypto currency? Either set up an entry wallet like Coinbase, an exchange like cross.io, or a mining pool?

-------------
Brian French
703-507-5383

mamchenkov:
I prefer `.gitkeep` to `.gitignore`, as `.gitignore` has another meaning, which might be confusing.

fr3nch13:
Or even better. Provide a codebase for such websites. 

-------------
Brian French
703-507-5383

mamchenkov:
@fr3nch13 well, we are a technology / software provider, when it comes to financial industry in general and forex in particular.  We don't have our own trading platforms or wallets, etc.  However, our products (Forex CRM, Forex Member's Area, Forex websites) integrate with third-party providers, which can easily include cryptocurrency tools and services.

Feel free to ping me at leonid@qobo.biz or via Skype at 'mamchenkov'. ;)

fr3nch13:
I meant ion that, because I’m like you, a coder first. But I like to dabble in that market, and there seems to be a void to fill by popping up sites for people that don’t want to do what we do. 

-------------
Brian French
703-507-5383

PR #481, opened 2018-02-26
Ajaxified system info tabs (task #5460)

anvyst:


src/Template/System/json/info.ctp
```diff
@@ -0,0 +1,5 @@
+<?php
+$element = 'System/' . $content;
+if ($this->elementExists($element)) {
+    echo $this->element('System/' . $content);
```
anvyst: @mamchenkov use `$element` inside `$this->element()` call?

PR #482, opened 2018-02-26
Many-to-many Modules script (task #5450)

georgeconstantinou:
The script handles migration of many-to-many associations to standalone Modules and removal of "manyToMany" section from the Modules `config.json` file.

**Requirements**
Merge and release of https://github.com/QoboLtd/cakephp-utils/pull/98

### DIFF
--- a/src/Shell/Task/Upgrade20180226Task.php
+++ b/src/Shell/Task/Upgrade20180226Task.php
@@ -1,0 +1,317 @@
+<?php
+namespace App\Shell\Task;
+
+use Cake\Console\Shell;
+use Cake\Core\Configure;
+use Cake\Filesystem\File;
+use Cake\Filesystem\Folder;
+use Cake\Utility\Inflector;
+use InvalidArgumentException;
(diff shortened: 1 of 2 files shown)


src/Shell/Task/Upgrade20180226Task.php
```diff
@@ -0,0 +1,317 @@
...
+
+        $this->success(sprintf('%s completed.', $this->getOptionParser()->getDescription()));
+    }
+
+    /**
+     * Validates CSV modules path.
+     *
+     * @return void
+     */
+    private function validatePath()
```
mamchenkov: @georgeconstantinou maybe we can use the [validatePath()`](https://github.com/QoboLtd/cakephp-utils/blob/master/src/Utility.php#L33) method from the Utils plugin.

georgeconstantinou: @mamchenkov done: https://github.com/QoboLtd/project-template-cakephp/pull/482/commits/647ad993c6cf8d09840186ba88b18529d030a732

PR #514, opened 2018-03-27
Added list of titles

stepanov:


stepanov:
@mamchenkov requested options are added

PR #570, opened 2018-07-18
Custom background images (task #6614)

anvyst:
By default we use `img/login/qobo` path to load random background images for `dark` skin.
In case you want to use your own background images, please add them into `img/login/custom` directory, and specify its name in `config/admin_lte.php` in `backgroundImages`


src/Template/Element/background-image-dark.ctp
```diff
@@ -0,0 +1,14 @@
+<?php
+use Cake\Core\Configure;
+use Cake\Filesystem\Folder;
+
+$backgroundImages = '/img/login/' . Configure::read('Theme.backgroundImages') . '/';
+
+$dir = new Folder(WWW_ROOT . $backgroundImages);
+$images = $dir->find();
+
+echo $this->Html->tag(
```
mamchenkov: @anvyst we should probably only output this if there is at least one image found.  Otherwise, the logs will be polluted with the 404 errors.

src/Template/Layout/AdminLTE/login.ctp
```diff
@@ -28,6 +28,11 @@ $this->Html->css('login-' . (string)(Configure::read('Theme.version') ?: 'light'
         <![endif]-->
     </head>
     <body class="hold-transition skin-<?php echo Configure::read('Theme.skin'); ?> login-page">
+        <?php
+        if ('dark' == Configure::read('Theme.version')) {
```
mamchenkov: @anvyst are we using the 'dark' theme anywhere else?  If not, we can probably get rid of this, as now we can configure different folders for the backgrounds.

anvyst: @mamchenkov , we use `dark` version of the login page, which is stated in `config/admin_lte.php` config file. With combination of dark/light template elements it works fine, when `Theme.version` (dark) is passed to elements as a suffix.

For example we have login-dark.ctp, and login-light.ctp. That's how AdminLTE solves the problem of color schema contrasting of dark/light backgrounds and fonts.

PR #571, opened 2018-07-18
Adding custom skins to template (task #6567)

anvyst:
For a quick customisation of the colour palettes for the template, we're using the same skin concept as AdminLTE. 

Custom skin will be stored as `skin-<name>.min.css` file located in `webroot/css/skins/`. 
It's a good practice to keep unminified version of the CSS file right next to the min file.

Example:
```php 

// config/admin_lte.php

'skin' => 'custom',
'skinUrl' => 'AdminLTE.skins/skin-blue.min', // example of plugin defined skin
'skinUrl' => 'skins/skin-custom', // definition of a local skin

```


config/admin_lte.php
```diff
@@ -21,6 +21,8 @@
             'show_social' => false,
         ],
         'version' => 'dark',
+        'skin' => 'blue',
+        'skinUrl' => 'AdminLTE.skins/skin-',
```
mamchenkov: @anvyst 

I think the `skinUrl` parameter should contain the full URL of the skin, as we discussed, not just the prefix.  This will allow to load skins from external URLs, other plugins, etc.  It will also allow to choose between the minified and non-minified version.

I don't think forcing the minification of the custom skin is a good thing.  It makes things more complicated (we currently don't have an automated toolchain to minify) and prone to human errors.

PR #576, opened 2018-07-23
Integration tests for swagger controllers (task #5711)

georgeconstantinou:


tests/TestCase/Controller/SwaggerUiControllerTest.php
```diff
@@ -0,0 +1,27 @@
...
+
+    /**
+     * Test index method
+     *
+     * @return void
+     */
+    public function testIndex()
+    {
+        if (! Plugin::loaded('Alt3/Swagger')) {
+            return;
```
mamchenkov: @georgeconstantinou I think the skipping of the test is better, as you can provide a reason, and it'll show the 'S' in the output, rather than a dot.

PR #583, opened 2018-07-24
Adding dynamic avatars (task #6433)

anvyst:
We introduce new concept of Avatars handling for the system.

Priority of avatar usage goes this way:

1. Your custom uploaded avatars will have higher priority over what you have
2. In case you have Gravatar account associated with the email. The system will pick it up
3. If non of the above worked - `DynamicAvatar` kicks in with your initials (based on email, name/username)
4. And for the others - using anonymous avatar.


src/Shell/AvatarsSyncShell.php
```diff
@@ -0,0 +1,71 @@
...
+        return $parser;
+    }
+
+    /**
+     * main() method.
+     *
+     * @return bool|int|null Success or error code.
+     */
+    public function main()
+    {
```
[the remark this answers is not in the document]
georgeconstantinou: @anvyst this can be easily fixed

src/Avatar/AvatarInterface.php
```diff
@@ -1,20 +1,52 @@
 <?php
 namespace App\Avatar;
 
-interface AvatarInterface
+use Cake\Core\Configure;
+
+abstract class AvatarInterface
```
georgeconstantinou: @anvyst is better to keep the Interface and create an AbstractAvatar class that implements the interface.

mamchenkov: @anvyst I'm not the biggest fan of this.  Interface is a reserved word and it has a particular meaning in programming in general, and in PHP in particular.  Abstract is a reserved word to, and 'abstract class' has a very specific meaning.  They are similar, but they are not the same thing.

I think the best approach is to have the `AvatarInterface` which defines the functionality.  And then, if you have several classes that share code, you can have the `abstrace class AbstractAvatar implements AvatarInterface`.  This will allow to reuse code, but without forcing all other implementations to inherit the code.

Given how varied are the sources for the user avatar, I think it's very likely that we'll have more providers, that will have to implement the interface, but won't have to share the code.

src/Avatar/AvatarInterface.php
```diff
@@ -1,20 +1,52 @@
...
      */
-    public function __construct(array $options);
+    public function processAvatarResource($file, $resource)
+    {
+        $processed = false;
+
+        imagealphablending($resource, false);
+        imagetruecolortopalette($resource, false, 256);
+
+        if (imagepng($resource, $file, 6, PNG_NO_FILTER)) {
```
mamchenkov: @anvyst can you please add a comment as to what is being done here and why.  It's not very obvious from the options here what's going on.

src/Avatar/Service.php
```diff
@@ -24,8 +39,62 @@ public function __construct(AvatarInterface $avatar)
      *
      * @return string
      */
-    public function getImage()
+    public function getImage(array $options)
     {
+        $filename = $this->getImageName($options);
+        $options['filename'] = $filename;
+
+        foreach (['ImageSource', 'Gravatar', 'DynamicAvatar'] as $classType) {
```
mamchenkov: @anvyst this should be in the configuration.

src/Avatar/Service.php
```diff
@@ -24,8 +39,62 @@ public function __construct(AvatarInterface $avatar)
...
+     * Initiate class object base on its name
+     *
+     * @param string $name of the class
+     * @param array $options if any required
+     *
+     * @return \App\Avatar\AvatarInterface $instance of the class
+     */
+    public function getAvatarSource($name, $options)
+    {
+        $class = __NAMESPACE__ . '\\Type\\' . $name;
```
mamchenkov: @anvyst this should also move to the configuration.  Providers can be outside of the current namespace.  The easiest is probably to use something like this in configuration:

```
'class' => \Some\Namespace\GoodAvatar::class,
```

src/Controller/UsersController.php
```diff
@@ -102,18 +105,29 @@ public function uploadImage($id)
             return $this->redirect($this->request->referer());
         }
 
-        $user = $this->Users->patchEntity($user, ['image' => $data]);
+        $extension = pathinfo($data['name'], PATHINFO_EXTENSION);
```
mamchenkov: @anvyst we should probably lowercase the extension as well, as it's used later on for comparisons.

src/Controller/UsersController.php
```diff
@@ -102,18 +105,29 @@ public function uploadImage($id)
...
+        if (in_array($extension, ['jpg', 'jpeg'])) {
+            $source = imagecreatefromjpeg($data['tmp_name']);
+        }
+
+        imagealphablending($source, false);
+        imagetruecolortopalette($source, false, 256);
+        $extension = Configure::read('Avatar.extension');
+
+        if (imagepng($source, $directory . $id . $extension, 6, PNG_NO_FILTER)) {
+            imagepng($source, $customDirectory . $id . $extension, 6, PNG_NO_FILTER);
```
mamchenkov: @anvyst `imagepng()` is called in multiple places with the same parameters (this class and in the `AbstractAvatar` as well).  This should be abstracted and reused.

src/Model/Entity/User.php
```diff
@@ -39,18 +38,17 @@ protected function _getName()
...
-        }
-
         $type = Configure::read('Avatar.default');
         $options = (array)Configure::read('Avatar.options.' . $type);
-        $options['email'] = (string)$this->get('email');
 
-        $service = new AvatarService(new $type($options));
+        $options = array_merge($options, [
+            'id' => $this->get('id'),
+            'email' => (string)$this->get('email'),
```
mamchenkov: @anvyst any particular reason why out of the three values only the email is cast to string?

src/Shell/AvatarsSyncShell.php
```diff
@@ -0,0 +1,68 @@
...
+        $avatarsPath = WWW_ROOT . $avatarsDir;
+        $customPath = WWW_ROOT . $customDir;
+
+        $usersCount = $query->count();
+        $generated = $updated = 0;
+
+        foreach ($query as $entity) {
+            $filename = $entity->id . '.png';
+
+            if (file_exists($customPath . $filename)) {
```
mamchenkov: @anvyst all the avatar updating logic should be in the Users model.  The shell should just call it for every user.

src/Shell/Task/Upgrade201807260843Task.php
```diff
@@ -0,0 +1,69 @@
...
+
+        foreach ($query->all() as $entity) {
+            $id = $entity->get('id');
+            $decodedImage = file_get_contents($entity->get('image'));
+            $source = imagecreatefromstring($decodedImage);
+
+            imagealphablending($source, false);
+            imagetruecolortopalette($source, false, 256);
+
+            if (imagepng($source, $directory . $id . $extension, 6, PNG_NO_FILTER)) {
```
mamchenkov: @anvyst another  `imagepng()` in here.

config/avatar.php
```diff
@@ -1,19 +1,29 @@
...
+            'App\Avatar\Type\Gravatar' => [
                 'size' => 160, // sets the desired image size
-                'default' => 'mm', // sets the default/fallback themed image
+                'default' => '404', // sets the default/fallback themed image
                 'rating' => 'g', // sets the desired image appropriateness rating
+            ],
+            'App\Avatar\Type\DynamicAvatar' => [
+                'size' => 160,
+                'length' => 2,
+                'background' => '#00c0ef',
```
mamchenkov: @anvyst can we also have several colors in here, and a random one picked during the avatar generation. [array_rand()](http://php.net/manual/en/function.array-rand.php) can easily handle that.

PR #586, opened 2018-07-31
Skip access check for Database lists (task #5578)

georgeconstantinou:


mamchenkov:
@georgeconstantinou we also need to remove the upgrade script which grants the permissions during the deploy.

PR #595, opened 2018-08-13
Menu listeners (task #6790) (task #6817)

softius:
Splits the existing Menu Listener into multiple menu listeners


src/Access/CapabilityTrait.php
```diff
@@ -0,0 +1,70 @@
...
+     */
+    protected function _checkAccess($url, $user)
+    {
+        $stringUrl = null;
+        if (is_array($url)) {
+            $stringUrl = Router::url($url);
+        } elseif (is_string($url)) {
+            $stringUrl = $url;
+        } else {
+            throw new \InvalidArgumentException();
```
mamchenkov: @softius please update with the use statement at the top, rather than namespaced classes.

src/Event/Plugin/Menu/View/ModuleMenuListener.php
```diff
@@ -0,0 +1,168 @@
...
+{
+    /**
+     * Menu item defaults.
+     *
+     * @var array
+     */
+    protected $defaults = [
+        'url' => '#',
+        'label' => 'Undefined',
+        'icon' => 'circle-o',
```
mamchenkov: @softius we use 'cube' as a default icon in a few places.  Whenever we see cube, we know we need to fix the icon. :)

PR #596, opened 2018-08-13
Improving unit tests (task #6846)

anvyst:
- Improving code coverage
- Added Abstract Scheduled Job class to simplify unit tests
- Used native Cake constants instead of `dirname()` 
- Added Cron Shell tests

### DIFF
--- a/src/Model/Table/ScheduledJobsTable.php
+++ b/src/Model/Table/ScheduledJobsTable.php
@@ -17,6 +17,8 @@ use RuntimeException;
 
 class ScheduledJobsTable extends AppTable
 {
+    const JOB_ACTIVE = 1;
+
     /**
      * Initialize method
      *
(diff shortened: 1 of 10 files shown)


src/Model/Table/ScheduledJobsTable.php
```diff
@@ -186,7 +203,7 @@ public function getList(array $options = [])
 
                 $result = array_merge($result, $object->getList());
             } catch (RuntimeException $e) {
-                pr($e->getMessage());
+                Log::warning($e->getMessage());
```
mamchenkov: @anvyst not having any handlers and throwing exception, is probably an error, not a warning.

src/Shell/CronShell.php
```diff
@@ -55,7 +55,7 @@ public function main()
         $this->ScheduledJobs = TableRegistry::get('ScheduledJobs');
         $this->ScheduledJobLogs = TableRegistry::get('ScheduledJobLogs');
 
-        $jobs = $this->ScheduledJobs->getActiveJobs();
+        $jobs = $this->ScheduledJobs->getJobs(true);
```
mamchenkov: @anvyst I think it should be `getJobs(ScheduledJobs::JOB_ACTIVE)` ... the parameter to the function is an integer, not a boolean.

src/Shell/CronShell.php
```diff
@@ -66,8 +66,12 @@ public function main()
         $now = Time::now();
 
         foreach ($jobs as $entity) {
+            $shouldRun = false;
+
             $rrule = $this->ScheduledJobs->getRRule($entity);
-            $shouldRun = $this->ScheduledJobs->timeToInvoke($now, $rrule);
+            if ($rrule instanceof \RRule\RRule) {
```
mamchenkov: @anvyst please move the namespace prefix to the use statement for clarity

tests/TestCase/Model/Table/ScheduledJobsTableTest.php
```diff
@@ -64,6 +64,15 @@ public function testGetActiveJobs()
         $this->assertInstanceOf('\Cake\ORM\ResultSet', $result);
     }
 
+    public function testGetJobs()
+    {
+        $result = $this->ScheduledJobsTable->getJobs(false);
```
mamchenkov: @anvyst parameter must be an integer, not a bool

PR #599, opened 2018-08-14
Admin menu access (task #6817)

softius:
Only superusers and users that belong to Admins group can have access to admin menu (task #6817)

### DIFF
--- a/src/Event/Plugin/Menu/View/ApplicationMenuListener.php
+++ b/src/Event/Plugin/Menu/View/ApplicationMenuListener.php
@@ -5,6 +5,8 @@ namespace App\Event\Plugin\Menu\View;
 use App\Access\CapabilityTrait;
 use Cake\Event\Event;
 use Cake\Event\EventListenerInterface;
+use Cake\ORM\TableRegistry;
+use Groups\Model\Table\GroupsTable;
 use Menu\Event\EventName as MenuEventName;
 use Menu\MenuBuilder\Menu;
 use Menu\MenuBuilder\MenuInterface;
@@ -48,7 +50,17 @@ class ApplicationMenuListener implements EventListenerInterface
 
         // Ugly hack to hide admin menu from non superusers
         if ($name == MENU_ADMIN) {
-            if (!$user['is_superuser']) {
+            $isSuperuser = $user['is_superuser'];
+
+            /** @var GroupsTable $groups */
+            $groups = TableRegistry::get('Groups.Groups');
+            $userGroups = $groups->getUserGroups($user['id']);
+            $inAdminGroup = in_array('Admins', $userGroups);
+
+            // User can have access to admin menu only and only if
+            // a) is a superuser
+            // b) belongs to Admins group
+            if (!$isSuperuser && !$inAdminGroup) {
                 $event->stopPropagation();
 
                 return;


src/Event/Plugin/Menu/View/ApplicationMenuListener.php
```diff
@@ -48,7 +50,17 @@ public function getMenuItems(Event $event, $name, array $user, $fullBaseUrl = fa
 
         // Ugly hack to hide admin menu from non superusers
         if ($name == MENU_ADMIN) {
-            if (!$user['is_superuser']) {
+            $isSuperuser = $user['is_superuser'];
+
+            /** @var GroupsTable $groups */
+            $groups = TableRegistry::get('Groups.Groups');
+            $userGroups = $groups->getUserGroups($user['id']);
+            $inAdminGroup = in_array('Admins', $userGroups);
```
mamchenkov: @softius OK, this works, but can be better. :) 

Firstly, I think the value of the group shouldn't be hardcoded in here, but used as a constant or a configuration.  We have some system groups configured in the Groups plugin, but they are defined as a list and a bit hard to fetch reliably.

For right now, I think setting the group name to a constant will work fine.

For the ideal world, I think we should rather rely on the role.  Currently, Admins group is assigned Admins role.  Admins role is not editable, so it is impossible to remove the Admins group from there.  But if in the future, we'll allow changing of roles in the Groups interface, there might be a change that won't notice.  Roles however, are more related to the access permissions, as they define the capabilities, and the Admins role is configured in the Roles and Capapbilities plugin and is easy to fetch.

I'll leave it up to you decide if we need to use the roles or stay with groups.  But if we stay with groups then at least please move the 'Admins' to a class constant.

softius: @mamchenkov you are right. this wasn't an elegant solution :)

src/Model/Entity/User.php
```diff
@@ -51,4 +54,32 @@ protected function _getImageSrc()
...
+        if ($this->get('is_superuser')) {
+            return true;
+        }
+
+        try {
+            /** @var CapabilitiesTable $capabilities */
+            $capabilities = TableRegistry::get('RolesCapabilities.Capabilities');
+            $userGroups = $capabilities->getUserGroups($this->get('id'));
+            $userRoles = $capabilities->getGroupsRoles($userGroups);
+            $isAdmin = in_array(Configure::read('RolesCapabilities.Roles.Admin.name'), $userRoles);
```
mamchenkov: @softius I think we can even use [readOrFail()](https://api.cakephp.org/3.4/class-Cake.Core.Configure.html#_readOrFail) method here, which will throw an exception if the value is not configured.  Just in case we move it somewhere else later.  Since the code is already wrapped in try/catch, the behavior will be the same, but it will avoid PHP errors with `in_array()`.

src/Event/Plugin/Menu/View/ApplicationMenuListener.php
```diff
@@ -46,13 +46,11 @@ public function getMenuItems(Event $event, $name, array $user, $fullBaseUrl = fa
             return;
         }
 
-        // Ugly hack to hide admin menu from non superusers
-        if ($name == MENU_ADMIN) {
-            if (!$user['is_superuser']) {
-                $event->stopPropagation();
+        // Only administrators can access the admin menu
+        if ($name === MENU_ADMIN && !$user['is_admin']) {
+            $event->stopPropagation();
```
mamchenkov: @softius any particular reason for stopping the propagation instead of just returning?

softius: Yes, since we want to stop menu generation. We don't want other listeners to proceed with generation. We can remove stopPropagation but then we will need to update all listeners and keep that in mind for any future implementations.

src/Model/Entity/User.php
softius: Sure we can do that.

src/Event/Plugin/Menu/View/ApplicationMenuListener.php
mamchenkov: @softius I think this is fine for now, since we only have one event anyway.  With multiple events, we'll need to control the order of the events as well, to ensure that this one executes first.

PR #600, opened 2018-08-14
Disable delete action when user is locked (task #6800)

softius:
A user can be deleted only and only if:

- User is not the system user defined in .env
- User is not the currently logged in user

### DIFF
--- a/src/Template/Users/index.ctp
+++ b/src/Template/Users/index.ctp
@@ -18,7 +18,7 @@ echo $this->Html->scriptBlock(
     '$(".table-datatable").DataTable({
         stateSave:true,
         paging:true,
-        searching:true
+        searching:false
     });',
     ['block' => 'scriptBottom']
 );
--- a/src/Template/Users/index.ctp
+++ b/src/Template/Users/index.ctp
@@ -94,16 +94,18 @@ echo $this->Html->scriptBlock(
                                     ['action' => 'change-user-password', $user->id],
                                     ['title' => __('Change User Password'), 'class' => 'btn btn-default btn-sm', 'escape' => false]
                                 ) ?>
-                                <?= $this->Form->postLink(
-                                    '<i class="fa fa-trash"></i>',
-                                    ['controller' => 'Users', 'action' => 'delete', $user->id],
-                                    [
-                                        'confirm' => __('Are you sure you want to delete # {0}?', $user->id),
-                                        'title' => __('Delete'),
-                                        'class' => 'btn btn-default btn-sm',
-                                        'escape' => false
-                                    ]
-                                ) ?>
+                                <?php if (!in_array($user->username, $lockedUsers)): ?>
+                                    <?= $this->Form->postLink(
+                                        '<i class="fa fa-trash"></i>',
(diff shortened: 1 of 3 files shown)


tests/TestCase/Controller/UsersControllerTest.php
```diff
@@ -320,28 +321,45 @@ public function testEditWithoutSession()
...
+    public function testDeleteSameUser()
+    {
+        $this->enableCsrfToken();
+        $this->enableSecurityToken();
+        $this->withSession();
+
+        try {
+            $this->disableErrorHandlerMiddleware();
+            $this->delete('/users/delete/' . $this->userId);
+            $this->fail('Expected ForbiddenException');
```
mamchenkov: @softius there is an easier way to test for expected exceptions.  See an example here:

https://github.com/QoboLtd/project-template-cakephp/blob/master/tests/TestCase/Feature/FactoryTest.php#L34

src/Template/Users/index.ctp
```diff
@@ -94,16 +94,30 @@ echo $this->Html->scriptBlock(
...
+                                            'disabled' => true,
+                                            'onClick' => 'return false',
+                                        ]
+                                    ) ?>
+                                <?php else: ?>
+                                    <?= $this->Form->postLink(
+                                        '<i class="fa fa-trash"></i>',
+                                        ['controller' => 'Users', 'action' => 'delete', $user->id],
+                                        [
+                                            'confirm' => __('Are you sure you want to delete # {0}?', $user->id),
```
mamchenkov: @softius while we are at it, I think we should show `$user->username` in the confirmation dialog, and not the `$user->id`.  It's more humane :)

tests/TestCase/Controller/UsersControllerTest.php
softius: Totally forgot about this! :+1:

src/Template/Users/index.ctp
softius: @mamchenkov We can also remove the data table search which is enabled for Users (for some reason)

mamchenkov: @softius there was a reason for the dataTables search, but I don't remember it top of my head.  Let's leave it for now.  Later I'll dig through the tickets and git history to see why it's there and if we need it at all. :)

src/Template/Users/index.ctp
```diff
@@ -94,16 +94,30 @@ echo $this->Html->scriptBlock(
...
-                                    '<i class="fa fa-trash"></i>',
-                                    ['controller' => 'Users', 'action' => 'delete', $user->id],
-                                    [
-                                        'confirm' => __('Are you sure you want to delete # {0}?', $user->id),
-                                        'title' => __('Delete'),
-                                        'class' => 'btn btn-default btn-sm',
-                                        'escape' => false
-                                    ]
-                                ) ?>
+                                <?php if (in_array($user->username, $lockedUsers)): ?>
```
anvyst: @softius I thinks it would be better if we simply hide delete button if the user cannot delete a record.

mamchenkov: I think it's fine ;)

PR #602, opened 2018-08-16
Retrieve only dashboards that user has access to (task #6867)

softius:


src/Event/Plugin/Menu/View/DashboardMenuListener.php
```diff
@@ -10,6 +10,7 @@
 use Menu\MenuBuilder\MenuInterface;
 use Menu\MenuBuilder\MenuItemContainerInterface;
 use Menu\MenuBuilder\MenuItemFactory;
+use Search\Model\Table\DashboardsTable;
```
anvyst: @softius you don't need to include the `use DashboardsTable` as it's loaded via `TableRegistry::get()`.

PR #622, opened 2018-09-05
Menu API & Listeners - Part 1 (task #809)

softius:
This PR is part of the work to be done for #809. Also, it depends on work completed by PR https://github.com/QoboLtd/cakephp-menu/pull/101

The most important changes is that menus are being constructed using event listeners for the following templates:

* src/Template/Element/Module/Menu/index_top.ctp
* src/Template/Element/menu-render.ctp
* src/Template/Plugin/Search/Element/Menu/search-view-actions.ctp

It also utilizes Menu API to minimize the boilerplate (such as rawHTML) as much as possible. There are two exceptions for this which are ‘Duplicates’ and ‘delete for data tables’ . I will try to provide an improved solution in another PR.

- Replace arrays with Menu API for search view actions Add Search View Menu Listener (task #809)
- Re-enable dynamic deletion for data tables (task #809)
- Include the displayField in confirm msg (task #809)
- Move index_top logic into ModuleIndexListener (task #809)
- Improve readability for script blocks (task #809)
- Move duplicates menu item in view element (task #809)
- Improve code readability (task #809)
- Improve code readability further (task #809)
- Use Module Index Listener (task #809)

PR #625, opened 2018-09-06
Menu API & Listeners - Part 2 (task #809)

softius:
This PR is part of the work to be done for #809. Also, it depends on work completed by PR QoboLtd/cakephp-menu#102

The most important changes is that menus are being constructed using event listeners for the following templates:
* src/Template/Plugin/Search/Element/Menu/dashboards-view-top.ctp
* src/Template/Element/Module/Menu/view_top.ctp

Also,
* Introduces a new Trait which facilitates Menu Constructions for common actions like edit, view and delete
* Utilizes new changes provided by Menu API to define elements instead of rawHTML. This improves the solution provided in #622


src/Event/Plugin/Menu/View/DashboardMenuListener.php
```diff
@@ -53,7 +72,7 @@ public function getMenuItems(Event $event, $name, array $user, $fullBaseUrl = fa
      * @param array $user Current user
      * @return void
      */
-    private function addDashboardItems(MenuInterface $menu, array $user)
+    private function addAdminMenuItems(MenuInterface $menu, array $user)
```
georgeconstantinou: @softius I think that the initial method name is more applicable.

softius: @georgeconstantinou method name was changed since Dashboard Listener handles both admin and dashboard view top menus, and addDashboardItems was a bit generic

PR #629, opened 2018-09-07
'Create dashboard' after dynamic login redirect (task #7039)

softius:
Improve handling of 'Create dashboard' after dynamic login redirect. In particular it will display the index page only and only if the no dashboards were found.

### DIFF
--- a/src/Template/Plugin/Search/Dashboards/index.ctp
+++ b/src/Template/Plugin/Search/Dashboards/index.ctp
@@ -18,13 +18,19 @@ use RolesCapabilities\Access\AccessFactory;
         <div class="col-xs-12">
             <div class="jumbotron">
                 <h1><?= __('Dashboards') ?></h1>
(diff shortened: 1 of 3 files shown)


src/Template/Plugin/Search/Dashboards/index.ctp
```diff
@@ -0,0 +1,32 @@
...
+
+use RolesCapabilities\Access\AccessFactory;
+
+?>
+<div class="container-fluid">
+    <div class="row">
+        <div class="col-xs-12">
+            <div class="jumbotron">
+                <h1><?= __('Dashboards') ?></h1>
+                <p><?= __('There are no configured Dashboards for you. Please contact the system administrator.') ?></p>
```
georgeconstantinou: @softius can we get rid of `Please contact the ...` if the user has access to create a dashboards?

PR #631, opened 2018-09-07
Database lists shell to upgrade task (task #5577)

georgeconstantinou:
Moved shell responsible for database lists creation to upgrade task.


mamchenkov:
@georgeconstantinou have a look at the CodeClimate issues ... some of them do indeed point to weird thingies :) 

https://codeclimate.com/github/QoboLtd/project-template-cakephp/pull/631

PR #641, opened 2018-09-21
Add shell scheduler garbage collector (task #7238)

eliamariutti:
Add the feature to clean the scheduled job logs from 
- the page /scheduled- job by button
- the shell `./bin/cake scheduled_log gc` (option -a "<time>")


src/Controller/ScheduledJobLogsController.php
```diff
@@ -22,6 +26,18 @@ public function index()
...
+     *
+     * Implement this method to avoid having to overwrite
+     * the constructor and call parent.
+     *
+     * @return void
+     */
+    public function initialize()
+    {
+        parent::initialize();
+    }
```
georgeconstantinou: @eliamariutti this method can be removed as it does not add any new functionality.

src/Event/Plugin/Menu/View/ModuleIndexListener.php
```diff
@@ -49,6 +49,10 @@ public function getMenuItems(Event $event, $name, array $user, $fullBaseUrl = fa
         $menu->addMenuItem($this->getImportMenuItem($request));
         $menu->addMenuItem($this->getAddMenuItem($request));
 
+        if ($request->param('controller') == 'ScheduledJobs') {
```
georgeconstantinou: @eliamariutti this can be moved in the `getDelLogItem` method as `disableIf()` callback (similar to line 80).

src/Shell/ScheduledLogShell.php
```diff
@@ -0,0 +1,86 @@
...
+     *
+     * @return array
+     */
+    protected function getDaysConfig()
+    {
+        $result = Configure::read('ScheduledLog.stats.age');
+        if (!$result) {
+            $this->info('Required parameter "age" is not defined (garbage collector)');
+
+            return;
```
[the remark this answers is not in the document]
georgeconstantinou: @eliamariutti I think the return value here should be a string.

tests/TestCase/Shell/ScheduledLogShellTest.php
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    /**
+     * Test getOptionParser method
+     *
+     * @return void
+     */
+    public function testGetOptionParser()
+    {
+        $this->markTestIncomplete('Not implemented yet.');
```
georgeconstantinou: @eliamariutti can you remove the empty test method to avoid too much noise in phpunit outputs.

tests/TestCase/Shell/ScheduledLogShellTest.php
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    /**
+     * Test main method
+     *
+     * @return void
+     */
+    public function testMain()
+    {
+        $this->markTestIncomplete('Not implemented yet.');
```
georgeconstantinou: @eliamariutti same here, can you remove the empty test method to avoid too much noise in phpunit outputs.

src/Shell/ScheduledLogShell.php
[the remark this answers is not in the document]
softius: @georgeconstantinou @eliamariutti I think we need to swap the checks. Check first if the parameter was provided and then fallback to Configuration.This way we can execute the command even if the Configuration does not exist

src/Template/ScheduledJobs/index.ctp
```diff
@@ -48,7 +48,9 @@ echo $this->Html->scriptBlock(
         <div class="col-xs-12 col-md-6">
             <div class="pull-right">
                 <div class="btn-group btn-group-sm" role="group">
-                    <?php echo $this->element('Module/Menu/index_top', ['user' => $user]);?>
+                    <?php
```
softius: @eliamariutti Please use one php short echo tags instead

PR #645, opened 2018-09-28
[WIP] Shell script populate DB with config (task #2048)

eliamariutti:
- first look how could be the structure of the admin_setting.php
- shell script that read chosen Configure from php files and put in the DB

### DIFF
--- a/config/Migrations/20180928131604_AdminSetting.php
+++ b/config/Migrations/20180928131604_AdminSetting.php
@@ -1,0 +1,34 @@
+<?php
+use Migrations\AbstractMigration;
+
+class AdminSetting extends AbstractMigration
+{
+    /**
+     * Change Method.
+     *
+     * More information on this method is available here:
+     * http://docs.phinx.org/en/latest/migrations.html#the-change-method
+     * @return void
+     */
+    public function change()
+    {
+        $table = $this->table('admin_settings');
+        $table->addColumn('key', 'string', [
+            'default' => null,
+            'null' => false,
+        ]);
+        $table->addColumn('value', 'string', [
+            'default' => null,
+            'limit' => 255,
+            'null' => false,
+        ]);
+        $table->addPrimaryKey([
+            'id',
+        ]);
+        $table->addColumn('timestamp', 'timestamp', [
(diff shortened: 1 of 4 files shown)


config/Migrations/20180928131604_AdminSetting.php
```diff
@@ -0,0 +1,34 @@
+<?php
+use Migrations\AbstractMigration;
+
+class AdminSetting extends AbstractMigration
```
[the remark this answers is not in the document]
softius: @eliamariutti please try to follow the naming conventions for Migrations. It should be something like `CreateAdminSettings` instead of `AdminSetting`

config/Migrations/20180928131604_AdminSetting.php
```diff
@@ -0,0 +1,34 @@
...
+     * Change Method.
+     *
+     * More information on this method is available here:
+     * http://docs.phinx.org/en/latest/migrations.html#the-change-method
+     * @return void
+     */
+    public function change()
+    {
+        $table = $this->table('admin_settings');
+        $table->addColumn('key', 'string', [
```
softius: @eliamariutti key must be unique, to avoid any surprises ;)

config/Migrations/20180928131604_AdminSetting.php
```diff
@@ -0,0 +1,34 @@
...
+            'default' => null,
+            'null' => false,
+        ]);
+        $table->addColumn('value', 'string', [
```
softius: @eliamariutti string might be too short in some cases, so we may need to change to large blob especially if we would want to store values like images (logo) etc

src/Shell/AdminSettingShell.php
```diff
@@ -0,0 +1,55 @@
...
+     *
+     * @return null Success or error code.
+     */
+    public function main()
+    {
+        $this->out('Populating the BD');
+
+        $settings = Configure::read('AdminSetting');
+
+        $query = TableRegistry::get('AdminSettings');
```
softius: @eliamariutti  Table and Entity class must be committed as well.

PR #646, opened 2018-10-02
Shell script populate DB with config (task #2048)

eliamariutti:
shell script that populate the DB
first idea of admin_setting.php file


src/Shell/AdminSettingShell.php
```diff
@@ -0,0 +1,55 @@
...
+     * @return null Success or error code.
+     */
+    public function main()
+    {
+        $this->out('Populating the BD');
+
+        $settings = Configure::read('AdminSetting');
+
+        $query = TableRegistry::get('AdminSettings');
+        $query->deleteAll([]);
```
georgeconstantinou: @eliamariutti I do not think the settings should be removed here unless they are read-only.

eliamariutti: Sorry @georgeconstantinou I forgot to mention that it was for debugging, just to check how the script was working.

config/Migrations/20180928131604_CreateSettings.php
```diff
@@ -0,0 +1,36 @@
...
+    {
+        $table = $this->table('settings');
+        $table->addColumn('key', 'string', [
+            'default' => null,
+            'null' => false,
+        ]);
+        $table->addIndex([
+            'key',
+        ], [
+            'name' => 'BY_KEY',
```
georgeconstantinou: @eliamariutti better to prefix the index name with _unique_, something like `unique_key`

src/Model/Table/SettingsTable.php
```diff
@@ -0,0 +1,64 @@
...
+    {
+        $validator
+            ->integer('id')
+            ->allowEmpty('id', 'create');
+
+        $validator
+            ->scalar('key')
+            ->maxLength('key', 255)
+            ->requirePresence('key', 'create')
+            ->notEmpty('key');
```
georgeconstantinou: @eliamariutti you can apply uniqueness validation here, see an example [here](https://github.com/QoboLtd/cakephp-cms/blob/master/src/Model/Table/SitesTable.php#L68).

src/Shell/SettingsShell.php
```diff
@@ -0,0 +1,84 @@
...
+     *
+     * @return null Success or error code.
+     */
+    public function main()
+    {
+        $this->out('Populating the DB');
+
+        $settings = Configure::read('Settings');
+        $query = TableRegistry::get('Settings');
+        foreach ($settings as $field => $data) {
```
softius: @eliamariutti to improve performance you can create an array including the `field` values which can be compared (diff) with all the defined aliases. This way you can attempt to insert only non existing keys into database.

src/Shell/SettingsShell.php
```diff
@@ -0,0 +1,84 @@
...
+            } catch (\Exception $e) {
+                if ($e->errorInfo[1] == 1062) {
+                    $this->out(sprintf("|%5.5s |%-40.40s | %-40.40s |", 'EXIST', $unit->key, $unit->value));
+                } else {
+                    $this->out($e);
+                }
+            }
+        }
+
+        return null;
```
softius: @eliamariutti we will need to return the actual exit code instead of always returning null.

src/Shell/SettingsShell.php
```diff
@@ -0,0 +1,84 @@
...
+        $settings = Configure::read('Settings');
+        $query = TableRegistry::get('Settings');
+        foreach ($settings as $field => $data) {
+            $unit = $query->newEntity([
+                'key' => $data['alias'],
+                'value' => Configure::read($data['alias']),
+            ]);
+
+            try {
+                $query->save($unit);
```
softius: @eliamariutti an alternative option is to use `saveMany` which "can save multiple entities atomically"

tests/Fixture/AdminSettingsFixture.php
```diff
@@ -0,0 +1,51 @@
+<?php
+namespace App\Test\Fixture;
```
softius: @eliamariutti I believe this file is no longer necessary

tests/Fixture/SettingsFixture.php
```diff
@@ -0,0 +1,52 @@
...
+    // @codingStandardsIgnoreEnd
+
+    /**
+     * Init method
+     *
+     * @return void
+     */
+    public function init()
+    {
+        $this->records = [
```
softius: @eliamariutti please update the fixture to include some real test data.

tests/TestCase/Model/Table/SettingsTableTest.php
```diff
@@ -0,0 +1,73 @@
...
+namespace App\Test\TestCase\Model\Table;
+
+use App\Model\Table\SettingsTable;
+use Cake\ORM\TableRegistry;
+use Cake\TestSuite\TestCase;
+
+/**
+ * App\Model\Table\SettingsTable Test Case
+ */
+class SettingsTableTest extends TestCase
```
softius: @eliamariutti I suggest to actually implement the test. Apart from other things, this will help you when you will be working on the next steps!

PR #659, opened 2018-10-09
Minor CS and import fixes

georgeconstantinou:


src/Event/Plugin/CsvMigrations/FieldHandlers/MagicDefaultValueListener.php
```diff
@@ -48,12 +49,11 @@ public function getDefaultValue(Event $event, $default = null)
         $magicValue = Inflector::camelize($magicValue);
         $magicValue = 'get' . $magicValue . 'Value';
 
-        // TODO: Add some logging here for non-supported magic values
         if (method_exists($this, $magicValue) && is_callable([$this, $magicValue])) {
-            $result = $this->$magicValue($fieldHandler);
+            return $this->$magicValue($fieldHandler);
         }
 
-        return $result;
```
mamchenkov: @georgeconstantinou the method should return what was passed, as a default, if nothing was done.  I think this line might prevent it from doing so.

PR #662, opened 2018-10-11
First concept of the UI for the settings (task #2048)

eliamariutti:
The UI is build on base of `config\settings.php`
The implementation of 'search' will be in (part n+1)
screenshot from 2018-10-12 10-44-47


src/Template/Settings/index.ctp
```diff
@@ -0,0 +1,136 @@
...
+<section class="content">
+	<div class="row">
+		<div class="col-md-12">
+			<div class="nav-tabs-custom">
+				<ul class="nav nav-tabs">
+					<?php
+					// Tab
+					$tabs = array_keys($data);
+					$first = true;
+					foreach ($tabs as $tab) {
```
softius: @eliamariutti please use the alternative syntax for control structures http://php.net/manual/en/control-structures.alternative-syntax.php

src/Template/Settings/index.ctp
```diff
@@ -0,0 +1,136 @@
...
+						$id_tab = str_replace(' ','_',$tab);
+						echo $first ? '<div class="tab-pane active" id="' . $id_tab . '">' : '<div class="tab-pane" id="' . $id_tab . '">';
+						?>
+						<div class="row">
+							<div class="col-md-3">
+								<ul class="nav nav-pills nav-stacked">
+									<?php
+									$columns = array_keys($data[$tab]);
```
softius: @eliamariutti I believe that we can drop the calls to array_keys, by using the `$key => $value` in `foreach` i.e. `foreach ($columns as $column => $columnValue)`

src/Template/Settings/index.ctp
```diff
@@ -0,0 +1,136 @@
...
+												'type' => $type,
+												'required' => true,
+											];
+											$inputField = $fhf->renderInput($field, $field, $value, ['fieldDefinitions' => $definition]);
+											echo $inputField;
+											if(isset($fieldValue['help']) ){
+												echo '<span class="help-block">'. $fieldValue['help'] .'</span>';
+											}
+										}
+										echo '</div>';
```
softius: @eliamariutti please try to avoid `echo 'html-goes-here`. You can use close / re-open PHP tags whenever necessary

PR #668, opened 2018-10-16
CakePHP App Upgrade (task #7909)

mamchenkov:
This PR is all about bringing our application to the latest version of the [CakePHP App](https://github.com/cakephp/app).

So far:

* Merge tag '3.0.0-RC2' (the closest to our starting point)
* Merge tag '3.0.0'
* Merge tag '3.0.3'
* Merge tag '3.0.4'
* Merge tag '3.0.5'
* Merge tag '3.1.0'
* Merge tag '3.2.0'

The rest will be merged as a separate PR, as [v3.2.1](https://github.com/cakephp/app/releases/tag/3.2.1) brings in the `.env` support via josegonzalez library, which conflicts with the vlucas one we already use.  Once we switch the libraries, we can continue with the merges of the CakePHP app skeleton.

SticklerCI comments will also be reviewed once we are up-to-date, if anything will remain.


config/bootstrap_cli.php
```diff
@@ -21,6 +21,14 @@
  * be put here.
  */
 
+// Set the fullBaseUrl to allow URLs to be generated in shell tasks.
+// This is useful when sending email from shells.
+//Configure::write('App.fullBaseUrl', php_uname('n'));
```
georgeconstantinou: @stepanov I think this functionality will be useful.

mamchenkov: @georgeconstantinou @stepanov we have our own in `config/bootstrap.php` which is more reliable.  We use our own ENV variables to get the correct URL, including the proper schema as well.

PR #677, opened 2018-10-26
Permission part 5 (task #2048)

eliamariutti:
- add new 'System settings' in the right menu
- `roles` in settings.php lists the access to the alias
- roles controls


src/Controller/SettingsController.php
```diff
@@ -24,25 +24,48 @@ class SettingsController extends AppController
      */
     public function index()
     {
-        $settings = $this->paginate($this->Settings);
-        $this->set(compact('settings'));
+        // Get User Roles
```
softius: @eliamariutti can we move the filtering part into the model and perhaps do some tests for that?

src/Controller/SettingsController.php
```diff
@@ -24,25 +24,48 @@ class SettingsController extends AppController
...
             $query = TableRegistry::get('Settings');
-            $type = Hash::combine(Configure::read('Settings'), '{s}.{s}.{s}.{s}.alias', '{s}.{s}.{s}.{s}.type');
+            $type = Hash::combine($dataFiltered, '{s}.{s}.{s}.{s}.alias', '{s}.{s}.{s}.{s}.type');
+            $roles = Hash::combine($dataFiltered, '{s}.{s}.{s}.{s}.alias', '{s}.{s}.{s}.{s}.roles');
 
             $set = [];
             foreach ($data as $key => $value) {
                 $entity = $query->findByKey($key)->firstOrFail();
+                // check the roles (never trust the user input)
+                if (count(array_intersect($roles[$key], $userRoles)) === 0) {
```
softius: @eliamariutti i think we can do this before even executing the query (so that we can return early)

src/Controller/SettingsController.php
```diff
@@ -24,25 +24,48 @@ class SettingsController extends AppController
...
-            throw new UnauthorizedException('Admin restricted area');
+        // Filter the Configure::read('Settings') with User Roles
+        $dataSettings = Configure::read('Settings');
+        $filter = array_filter(Hash::flatten($dataSettings), function ($value) use ($userRoles) {
+            return in_array($value, $userRoles);
+        });
+        $dataFlatten = [];
+        foreach ($filter as $key => $value) {
+            $p = explode('.', $key);
+            $p = $p[0] . '.' . $p[1] . '.' . $p[2] . '.' . $p[3];
```
georgeconstantinou: @eliamariutti these indexes might not always be available.

eliamariutti: @georgeconstantinou I see your point, but right now the `settings.php` has a structure of fix 4 levels and the flatten array must have all the index. If it's a "broke" configuration you will have much more troubles before reaching this point.

mamchenkov: @eliamariutti ,

it's all about [Defensive Programming](https://en.wikipedia.org/wiki/Defensive_programming).  Irrelevant of other issues, this code relies on things which might not be true.  It should check that they are true and either fail or fallback gracefully.

softius: @georgeconstantinou @mamchenkov can you please update your review?

PR #684, opened 2018-11-05
Bootstrap change order (part 8) (task #2048)

eliamariutti:
- Small changes in the order of what is loaded in bootstrap.
- The new `DbConfig` will provide some custom settings from the DB after all the php config files are loaded, and overwrite their configuration. 
- Every `Configure::load()` after the new DbConfig, will overwrite the custom settings.


config/bootstrap.php
```diff
@@ -87,17 +87,36 @@
...
+ConnectionManager::setConfig(Configure::consume('Datasources'));
+
+/**
+ * Load custom settings from the DB
+ */
+try {
+    Configure::config('dbconfig', new DbConfig());
+    Configure::load('Settings', 'dbconfig', true);
+} catch (\Cake\Database\Exception $e) {
+    // Do nothing
```
mamchenkov: @eliamariutti @softius I'm not very sure this is a good idea. :)

softius: @mamchenkov this was a workaround to avoid the chicken-egg problem i.e. when the table hasn't created yet but we need the application bootstrap to be executed without errors so that we can create the 'settings' table. Definetely can be improved though. @eliamariutti i recall that you try to check if the table exists at some point. can we revisit that?

src/Settings/DbConfig.php
```diff
@@ -10,12 +10,18 @@ class DbConfig implements ConfigEngineInterface
...
      */
     public function read($key)
     {
         $query = TableRegistry::get($key);
-        $data = $query->find('list', ['keyField' => 'key', 'valueField' => 'value'])->toArray();
+
+        // if the table $key doesn't exist will merge an empty array
+        try {
+            $data = $query->find('list', ['keyField' => 'key', 'valueField' => 'value'])->toArray();
+        } catch (\Exception $e) {
```
anvyst: @eliamariutti won't it be better to declare `use Exception` on top of the file? As in line 22 you backslash Exception thrown, but in `dump()` method you don't do it. Just for consistency reasons.

PR #685, opened 2018-11-07
Scope - part 9 (task #2048)

eliamariutti:
in the SettingsController :
`app()` will set App level settings
`user(uuid)` will allow to change settings to any particular user
`my()` will change settings of logged user

to keep in mind, `user` settings will be loaded at the begin of all controllers and overwrite `app` settings.


src/Model/Table/SettingsTable.php
```diff
@@ -93,31 +99,35 @@ public function settingsValidator($value, $context)
...
      * @throws \RuntimeException when settings.php structure is broke
      */
-    public function filterSettings($dataSettings, $userRoles)
+    public function filterSettings($dataSettings, $userScope)
     {
-        $filter = array_filter(Hash::flatten($dataSettings), function ($value) use ($userRoles) {
-            return in_array($value, $userRoles);
+        // debug(Hash::flatten($dataSettings));
+        $filter = array_filter(Hash::flatten($dataSettings), function ($value) use ($userScope) {
+            // foreach ($value as $key ) {
```
softius: @eliamariutti remove commented out code

src/Event/Component/SettingsListener.php
```diff
@@ -0,0 +1,47 @@
...
+            $userId = $event->getSubject()->Auth->user('id');
+        } catch (\Error $e) {
+            return;
+        }
+
+        if (empty($userId)) {
+            return;
+        }
+
+        try {
```
softius: @eliamariutti we dont catch the Exception in bootstrap. Is there any reason to catch the Exception here?

src/Controller/SettingsController.php
```diff
@@ -72,32 +154,34 @@ public function index()
...
         // For seach the new fields to insert
         $data = Hash::flatten(Configure::read());
         $this->set('alldata', $data);
 
-        // list of roles
-        $capabilities = TableRegistry::get('QoboRoles')->find('list', ['keyField' => 'name'])->toArray();
-        $this->set('roles', array_keys($capabilities));
-    }
+        // list of scope
+        $this->set('scope', ['app', 'user']);
```
softius: @eliamariutti we can define the available scopes as constants in Table class so that we avoid using 'app' and 'user' as hardcoded values

src/Controller/SettingsController.php
```diff
@@ -72,32 +154,34 @@ public function index()
...
-     * @return null Nothing to return
-     */
-    public function returnArray()
-    {
-        $this->autoRender = false;
         if ($this->request->is('post')) {
-            var_export($this->request->data());
+            $this->autoRender = false;
+
+            return var_export($this->request->data());
```
softius: @eliamariutti by default `var_export` outputs the variable representation, so either we need to omit the return or change the call parameters.

config/bootstrap.php
```diff
@@ -274,8 +278,6 @@
  * before any of our plugins that use routes, it breaks
  * them, needs to be investigated further.
  */
-Configure::write('Users.config', ['users']);
```
georgeconstantinou: @eliamariutti is the todo comment here resolved? Because according to it, loading the Users plugin before the rest of the plugins was breaking the routes. If this is no longer the case, we can get rid of the todo comment.

src/Controller/SettingsController.php
```diff
@@ -17,53 +18,168 @@
...
+            $scope = Hash::combine($dataFiltered, '{s}.{s}.{s}.{s}.alias', '{s}.{s}.{s}.{s}.scope');
 
             $set = [];
             foreach ($dataPut as $key => $value) {
-                // check the roles (never trust the user input)
-                if (count(array_intersect($roles[$key], $userRoles)) === 0) {
-                    $this->Flash->error(__('Failed to update settings, please try again.'));
-                    throw new UnauthorizedException('Can not update');
+                // if the key doesn't exist it fails.
+                $entity = $this->query->findByKey($key)->firstOrFail();
```
georgeconstantinou: @eliamariutti is this line required?

src/Template/Settings/index.ctp
```diff
@@ -85,7 +85,8 @@ $(document).ready(function(){
 										<?php
 										// Fields
 										foreach ($fields as $field => $fieldValue) :
-											$value = Configure::read($fieldValue['alias']);
+											// $value = Configure::read($fieldValue['alias']);
```
georgeconstantinou: @eliamariutti can you remove this.

src/Event/Component/SettingsListener.php
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * Load User settings and overwrite defaults
+     *
+     * @param \Cake\Event\Event $event Event instance
+     * @return void
+     */
+    public function loadUserSettings(Event $event)
+    {
+        try {
```
georgeconstantinou: @eliamariutti instead of a try...catch here you can check if the controller has the relevant component loaded. Something like:
```php
if (! $event->getSubject()->components()->has('Auth')) {
    return;
}
```

config/bootstrap.php
softius: @georgeconstantinou we didn't encountered any issues. I believe this was solved once we manage to resolve the conflict between `CakeDC` and `App`. Nevertheless, @eliamariutti can you please double check that routes are OK (especially those under `/users/` path)

config/settings.php
```diff
@@ -6,56 +6,64 @@
       'UI' => [
         'Theme' => [
           'Title' => [
-            'alias' => 'Theme.title',
+            'alias' => 'ScheduledLog.stats.age',
```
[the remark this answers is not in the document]
softius: @georgeconstantinou this is an alias for a Configuration key (parameter for `Configure::read` which already exists.

src/Event/Component/SettingsListener.php
softius: Nice :+1:

config/bootstrap.php
eliamariutti: @georgeconstantinou @softius I moved up trying to resolve some other issues, but since were not correlated with it, I can put it back as last plugin load.
Anywas, now the plugin is working great, there are no problems to load it first, at least on the `project`.

georgeconstantinou: @eliamariutti no need to revert this, plugin loading order should not affect routes in the first place.

config/settings.php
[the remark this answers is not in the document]
georgeconstantinou: @softius I understand this, but it seems like this is the Theme section of the settings, how is scheduled log related?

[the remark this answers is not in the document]
softius: @georgeconstantinou you are right it's not! :) the contents of this file are kinda temporary, with random configuration keys at the moment, since it's not clear what the final `settings.php` will include (meaning the actual configuration entries that it will include).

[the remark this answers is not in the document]
georgeconstantinou: @softius ok got it.

PR #686, opened 2018-11-07
CronShell fix (task #7968)

georgeconstantinou:
Replace slashes with underscores when generating lock files for plugin shell tasks with a vendor prefix. Otherwise, lock-file will be treated as sub-directory and will always fail to execute the relevant task.

### DIFF
--- a/src/Shell/CronShell.php
+++ b/src/Shell/CronShell.php
@@ -118,7 +118,7 @@ class CronShell extends Shell
      */
     public function lock(string $file, string $class)
     {
-        $class = str_replace([':', '/'], ['_', '_'], $class);
+        $class = preg_replace('/[^\da-z]/i', '_', $class);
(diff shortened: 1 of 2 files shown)


src/Shell/CronShell.php
```diff
@@ -118,7 +118,7 @@ public function main()
      */
     public function lock(string $file, string $class)
     {
-        $class = str_replace(':', '_', $class);
+        $class = str_replace([':', '/'], ['_', '_'], $class);
```
mamchenkov: @georgeconstantinou I think we should make it even more rigorous.  Instead of trying to match a particular character (we need to add backslashes, spaces, dollar signs, exclamation and question marks, etc), we should replace ALL non-alphanumeric characters with an underscore.

mamchenkov:
@georgeconstantinou approved, but ideally it'd be good to have the unit tests for this as well ;)

PR #687, opened 2018-11-08
Add search to logs (Task #5037)

mitsios:
Enabled search for logs. I had to create another table Logs that extends the existing DatabaseLogs in order for the search to work


src/Model/Entity/Log.php
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     * Fields that can be mass assigned using newEntity() or patchEntity().
+     *
+     * Note that when '*' is set to true, this allows all unspecified fields to
+     * be mass assigned. For security purposes, it is advised to set '*' to false
+     * (or remove it), and explicitly make individual fields accessible as needed.
+     *
+     * @var array
+     */
+    protected $_accessible = [
```
anvyst: @mitsios , I think you can minimise the list of fields by a short cut as in other `Entity` classes like:

```php
    protected $_accessible = [
        '*' => true,
    ];
```

src/Model/Table/LogsTable.php
```diff
@@ -0,0 +1,98 @@
...
+class LogsTable extends DatabaseLogsTable
+{
+
+    /**
+     * Initialize method
+     *
+     * @param array $config The configuration for the Table.
+     * @return void
+     */
+    public function initialize(array $config)
```
georgeconstantinou: @mitsios this might not be required if there are no changes from the parent method.

src/Model/Entity/Log.php
mitsios: Indeed, i changed the code to the one you suggested

src/Model/Table/LogsTable.php
mitsios: The only issue here is the table name that we need to use the same as the DatabaseLogs, i just removed some extra lines and i left the part for configuring the table name to DatabaseLogs instead of Logs

softius: @mitsios similary we can remove `validationDefault` and `use Cake\Validation\Validator;`

src/Model/Table/LogsTable.php
```diff
@@ -0,0 +1,94 @@
+<?php
+namespace App\Model\Table;
+
+use Cake\Validation\Validator;
+use DatabaseLog\Model\Table\DatabaseLogsTable;
+
+/**
+ * DatabaseLogs Model
```
softius: @mitsios must be `Logs Model` instead

src/Model/Entity/Log.php
```diff
@@ -0,0 +1,36 @@
+<?php
+namespace App\Model\Entity;
+
+use DatabaseLog\Model\Entity\DatabaseLog;
+
+/**
+ * DatabaseLog Entity
```
softius: @mitsios must be `Logs Entity` instead

src/Controller/LogsController.php
```diff
@@ -3,6 +3,7 @@
 
 use Cake\Core\Configure;
 use Cake\I18n\Time;
+use Cake\ORM\TableRegistry;
```
georgeconstantinou: @mitsios this class is not used anywhere.

src/Controller/LogsController.php
```diff
@@ -35,6 +36,7 @@ public function initialize()
         parent::initialize();
         $this->paginate['limit'] = 10;
         $this->paginate['fields'] = null;
+        $this->modelClass = 'DatabaseLog.DatabaseLogs';
```
georgeconstantinou: I think this is not required with the latest release of cakephp-search plugin, which includes the following fix: https://github.com/QoboLtd/cakephp-search/pull/224

PR #688, opened 2018-11-08
Multi aliases (part10) (task #2048)

eliamariutti:
Assumed that an alias that trigger another one is a special case, and change `alias` in `settings.php` from a string to an array is a long process that will compromise most of the work done till now, we have introduce the array `links` in `settings.php`. Any key in this array, will be a new record in the DB with the same value of the alias.


src/Model/Table/SettingsTable.php
```diff
@@ -144,6 +144,8 @@ public function valueValidator($value, $context)
      */
     public function createEntity($key, $value, $type, $scope, $context)
     {
+        // if the key doesn't exist it fails.
+        $entity = $this->findByKey($key)->firstOrFail();
```
eliamariutti: @georgeconstantinou  I think that now this line here is needed, since I moved "everything" in the model.

src/Controller/SettingsController.php
```diff
@@ -139,33 +158,21 @@ private function settings()
...
-                $entity = $this->query->find('all')->where(['key' => $key, 'scope' => $this->scope, 'context' => $this->context])->first();
-
-                // will storage only the modified settings
-                if (!is_null($entity) && $entity->value === $value) {
-                    // if the user setting match the app setting, the entity will be deleted
-                    if ($this->scope === SettingsTable::SCOPE_USER && $value === $this->dataApp[$key]) {
-                        $this->query->delete($entity);
-                    }
+                $entity = $this->query->createEntity($key, $value, $type[$key], $this->scope, $this->context);
+                !empty($entity) ? ($set[] = $entity) : '';
```
softius: @eliamariutti please avoid using ternary operators to assign a value. Try to use an `if` statement in such cases.

src/Controller/SettingsController.php
```diff
@@ -139,33 +158,21 @@ private function settings()
...
-                    // dynamic field to pass type to the validator
-                    'type' => $type[$key]
-                ];
-
-                // if (entity not exist) : new ? patch
-                $newEntity = is_null($entity) ? $this->Settings->newEntity($params) : $this->Settings->patchEntity($entity, $params);
-                $set[] = $newEntity;
+                foreach ($links[$key] as $link => $keyLink) {
+                    $entity = $this->query->createEntity($keyLink, $value, $type[$key], $this->scope, $this->context);
+                    !empty($entity) ? ($set[] = $entity) : '';
```
softius: @eliamariutti same as above

src/Model/Table/SettingsTable.php
```diff
@@ -133,6 +133,47 @@ public function valueValidator($value, $context)
...
+        $params = [
+            'key' => $key,
+            'value' => $value,
+            'scope' => $scope,
+            'context' => $context,
+            // dynamic field to pass type to the validator
+            'type' => $type
+        ];
+
+        // if (entity not exist) : new ? patch
```
softius: @eliamariutti bit confused since an exception will be thrown if the key does not exist (first method line)

src/Shell/SettingsShell.php
```diff
@@ -75,15 +80,41 @@ public function main()
...
+    }
+
     /**
      * reset() method. Truncate table Settings
      *
+     * @param string $key key of DB to delete
      * @return void
      */
-    public function reset()
+    public function reset($key = '')
```
softius: @eliamariutti I would prefer to have a `resetAll` method to avoid accidental truncation of the settings table

src/Controller/SettingsController.php
```diff
@@ -87,6 +87,9 @@ public function user($context)
         $this->dataSettings = Hash::merge($this->dataSettings, Hash::expand($this->dataApp), Hash::expand($dataUser));
         $this->viewBuilder()->template('index');
 
+        $userName = TableRegistry::get('Users')->find('list')->where(['id' => $context])->toArray();
+        $this->set('afterTitle', ' » ' . $userName[$context]);
```
softius: @eliamariutti can we move the double arrow into the ctp?

src/Shell/SettingsShell.php
```diff
@@ -48,17 +58,22 @@ public function main()
...
-                    'context' => 'app',
-                    'type' => $alias[$set] // dynamic field to pass `type` to the validator
-                ];
+            $data[] = $this->setData($alias, $set, $set);
+            if (empty($links[$set])) {
+                continue;
+            }
+            foreach ($links[$set] as $aliases => $value) {
+                if (in_array($value, $settings)) {
+                    throw new \Exception('Duble alias found');
```
softius: @eliamariutti minor typo

src/Model/Table/SettingsTable.php
```diff
@@ -133,6 +133,48 @@ public function valueValidator($value, $context)
...
+        // It will check if there is any record with a key = $key.
+        // if doesn't, it means that Settings table is not updated with settings.php.
+        $this->findByKey($key)->firstOrFail();
+        // select based on key, scope, conext
+        $entity = $this->find('all')->where(['key' => $key, 'scope' => $scope, 'context' => $context])->first();
+
+        // will storage only the modified settings
+        if (!is_null($entity) && $entity->value === $value) {
+            // if the user setting match the app setting, the entity will be deleted
+            if ($scope === 'user' && $value === $this->dataApp[$key]) {
```
softius: @eliamariutti you can use `SettingsTable` constants instead

src/Model/Table/SettingsTable.php
```diff
@@ -133,6 +133,48 @@ public function valueValidator($value, $context)
...
+     * @param  string $type  type
+     * @param  string $scope app, user, os?, env?
+     * @param  string $context uuid, value
+     * @return \App\Model\Entity\Setting|void
+     */
+    public function createEntity($key, $value, $type, $scope, $context)
+    {
+        // It will check if there is any record with a key = $key.
+        // if doesn't, it means that Settings table is not updated with settings.php.
+        $this->findByKey($key)->firstOrFail();
```
georgeconstantinou: @eliamariutti `firstOrFail()` will throw an exception here, is this the expected behavior?

PR #693, opened 2018-11-16
Fix and Features - part 11 (task #2048)

eliamariutti:
ezgif com-video-to-gif

- Add jquery search in settings
- Small fix and improvements


src/Controller/SettingsController.php
```diff
@@ -210,9 +210,12 @@ public function generator()
 
         // For render the main structure
         $dataSettings = Configure::read('Settings');
-        $this->set('data', $dataSettings);
+        $this->set('data', empty($dataSettings) ? null : $dataSettings);
```
softius: @eliamariutti since we are using `foreach` to iterate in view template, it's better to assign an empty array (to avoid errors that `$data` is not iterable. Also default values can be assigned using something like `Configure::read('Settings', []);`

config/settings.php
```diff
@@ -2,110 +2,61 @@
...
-        ],
-        'Logo_(large)' => [
-          'alias' => 'Theme.logo.large',
-          'links' => '',
-          'type' => 'string',
-          'scope' => [
-            (int)0 => 'app'
-          ]
-        ],
+  'first' => [
```
softius: @eliamariutti I suggest to replace `settings.php` with an empty array and perhaps a commented-out example, until we figure out what will be included.

PR #698, opened 2018-11-20
Shell script trash (task #2555)

eliamariutti:
The shell script will check in all tables that has the `Trash` behavior, if there is any trashed records and remove those older than `Configure::read('Trash.stats.age')` or by parameter `-a '<age>'`
`./bin/cake trash clear <module> -a '-1 day' --dry-run`
`./bin/cake trash clear_all -a '-1 day' --dry-run`
`./bin/bake trash list_modules`


src/Shell/TrashShell.php
```diff
@@ -0,0 +1,169 @@
...
+
+        $query = TableRegistry::get($module);
+
+        if (!$query->behaviors()->has('Trash')) {
+            // Module don't support trash
+            return;
+        }
+
+        try {
+            $trashEntities = TableRegistry::get($module)->find('onlyTrashed')->toArray();
```
softius: @eliamariutti instead of retrieving the whole dataset we can only count the records. Suggestion follows:
```suggestion
            $query = TableRegistry::get($module)->find('onlyTrashed');
            $total = $query->select(['total' => $query->func()->count('*')])->first()->toArray();
```

Also the above should be executed only and only if we are running in a `dry-run` mode (since we are not using `$trashEntities` anywhere else.

src/Shell/TrashShell.php
```diff
@@ -0,0 +1,163 @@
...
+    /**
+     * Get the option parser for this shell.
+     *
+     * @return \Cake\Console\ConsoleOptionParser
+     */
+    public function getOptionParser()
+    {
+        $parser = parent::getOptionParser();
+        $option = [
+                    'age' => [
```
georgeconstantinou: @eliamariutti indentation looks off here.

PR #699, opened 2018-11-20
Admin settings (task #2048)

eliamariutti:
**Admin settings** provides a way to storage part of the configuration of the application in the database and define custom configuration for each user. Here, a list of the main features : 

- **bootstrap** : The new `DbConfig` provides the custom application values loaded from the DB and overwrite those loaded from php files. Every `Configure::load(<some values>)` after the new DbConfig, will overwrite the custom settings.

- **settings.php** : the file contains the list of variables stored in the DB and the structure for the view `/settings`
The `Settings` array has four levels that build the view. Each keys refer to :
    `0`. name of the `tab`
    `1`. name of the `column`
    `2`. name of `section`
    `3`. name of the `title`
The `title` array keys are:
`alias` : application variable name that you want to store in the DB (`Configure::read()` arguments )
`type` : type for validation and UI input. Could be `integer`, `string` and `boolean`. Array and Objects are not supported yet.
`scope`: implemented scopes are ‘app’ and ‘user’. If it sets as ‘app’ only an administrator could write it. If set as ‘user’ it must have ‘app’ as well. `user` settings will overwrite `app` settings. Each user can read/write their own settings.
`link`: array where each value will provide a new record in the DB with the some value of the alias.
`help`: label displayed in `/settings` under each alias that helps users to understand the setting.

**NB** Each time that `settings.php` is modified, the database have to updated by shell script.

- **`http://localhost:<port>/settings/generator`**: Build the `settings.php` manually could be complicated. This interface only for developer in localhost, will provide a way to generate the array `Settings` in `settings.php`.

- **settings shell** : the command `./bin/cake settings` will populate the table `settings` according to the file `settings.php`. It is executed during the `./bin/build app:install` and `./bin/build app:update`
    `./bin/cake settings resetAll` will truncate the table
    `./bin/cake settings reset <key>` will delete all records with the key = 'key'.


softius:
@eliamariutti it looks like the build fails, with errors coming from phpunit. Can you please check ?

src/Template/Settings/generator.ctp
```diff
@@ -0,0 +1,343 @@
...
+    ['block' => 'scriptBottom']
+);
+
+$this->Html->scriptStart(array('block' => 'scriptBottom', 'inline' => false)); ?>
+
+$(document).ready(function(){
+    // settings.php array in json
+    let data = (<?= json_encode($data) ?>)
+    // Tabs
+    $.each(data,function(tab, cols){
```
anvyst: 4 nested `$.each` loops. Can we refactor it?

config/Migrations/20180928131604_CreateSettings.php
```diff
@@ -0,0 +1,40 @@
...
+     * Change Method.
+     *
+     * More information on this method is available here:
+     * http://docs.phinx.org/en/latest/migrations.html#the-change-method
+     * @return void
+     */
+    public function change()
+    {
+        $table = $this->table('settings');
+        $table->addColumn('key', 'string', [
```
anvyst: @eliamariutti throughout the code we often refer to settings by its `key`. What if we add `key` as table index additionally to `id`?

src/Controller/SettingsController.php
```diff
@@ -0,0 +1,255 @@
...
+    public function initialize()
+    {
+        parent::initialize();
+        $this->dataSettings = Configure::read('Settings');
+        /**
+         * @var \App\Model\Table\SettingsTable $table
+         */
+        $table = TableRegistry::get('Settings');
+        $this->query = $table;
+        $this->dataApp = $this->query->find('list', ['keyField' => 'key', 'valueField' => 'value'])
```
anvyst: @eliamariutti similar query find() statement appears 3 times in the controller, but with different conditions. Shall we move it to Settings Table as a method?

src/Model/Table/SettingsTable.php
```diff
@@ -0,0 +1,239 @@
...
+     * @return \App\Model\Entity\Setting|void
+     */
+    public function createEntity(string $key, string $value, string $type, string $scope, string $context)
+    {
+        // It will check if there is any record with a key = $key.
+        // if doesn't, it means that Settings table is not updated with settings.php.
+        $this->find('all')->where(['key' => $key])->firstOrFail();
+        // select based on key, scope, conext
+        $entity = $this->find('all')->where(['key' => $key, 'scope' => $scope, 'context' => $context])->first();
+        // Manly need for phpstan
```
anvyst: typo in the comment

src/Template/Settings/generator.ctp
eliamariutti: @anvyst It works exactly like the `src/Template/Settings/index.ctp`, with 4 nested `for` loops.
I can understand that is not the best solution but this is a tools only for developers to help to build/edit the very complex structure of the 4 dimensions array and it does only generate a `<div>` with some text in it.

src/Controller/SettingsController.php
eliamariutti: @anvyst I add a custom finder

config/Migrations/20180928131604_CreateSettings.php
softius: @anvyst the point is valid. We will address this later on.

PR #702, opened 2018-11-21
Added a controller and page to display the tables that support the trash behavior (task #2555)

mitsios:
Add a new controller and page to display the tables that support 'Trash'

By clicking the table name it will redirect to advance search with already having as filter the trashed filed with the option "greater" than 1900-01-01


src/Event/Plugin/Search/Model/SearchableFieldsListener.php
```diff
@@ -72,7 +72,7 @@ public static function getSearchableFieldsByTable(RepositoryInterface $table, ar
             return $result;
         }
 
-        if ($table->hasField('trashed')) {
+        if ($user['is_admin'] && $table->hasField('trashed') && !empty($fields)) {
```
georgeconstantinou: @mitsios no need for the `empty($fields)` check, is done above on line 71.

src/Template/Trash/index.ctp
```diff
@@ -0,0 +1,62 @@
...
...d="POST"><div style="display:none;"><input type="hidden" name="_method" value="POST" /><input type="hidden" name="_csrfToken" autocomplete="off" value="<?= $this->request->param('_csrfToken') ?>" /></div><div class="form-group search-field-wrapper"><input type="hidden" name="criteria[' + $(this).attr('data-table') + '.trashed][0][type]" value="datetime"><div class="row"><div class="col-xs-12 col-md-3 col-lg-2"><label>Trashed</label></div><div class="col-xs-4 col-md-2 col-lg-3"><select name="criteria[' + $(this).attr('data-table') + '.trashed][0][operator]" class="form-control input-sm"><option value="is">is</option><option value="is_not">is not</option><option value="greater" selected="">from</option><option value="less">to</option></select></div><div class="col-xs-6 col-md-5 col-lg-4"><div class="form-group input text"><div class="input-group input-group-sm "><div class="input-group-addon"><i class="fa fa-calendar"></i></div><input type="text" name="criteria[' + $(this).attr('data-table') + '.trashed][0][value]" data-provide="datetimepicker" data-magic-value="1" autocomplete="off" class="form-control input-sm" id="" value="1900-01-01 00:00"></div></div></div></div></div></form>');
```
georgeconstantinou: @mitsios I guess you can remove any styling elements from the form, such as div .row and .col-*.

tests/Fixture/ThingsTable.php
```diff
@@ -29,6 +29,13 @@ public function getFieldsDefinitions(array $stubFields = [])
                 'required' => null,
                 'non-searchable' => 1,
                 'unique' => null,
+            ],
+            'trashed' => [
```
softius: @mitsios :+1:

src/Controller/TrashsController.php
```diff
@@ -0,0 +1,36 @@
+<?php
+namespace App\Controller;
+
+use App\Controller\AppController;
+use Cake\Datasource\ConnectionManager;
+use Cake\ORM\TableRegistry;
+use Cake\Routing\Router;
```
softius: @mitsios `Cake\Routing\Route` is not used

src/Controller/TrashsController.php
```diff
@@ -0,0 +1,36 @@
...
+{
+    /**
+     * Display the tables that support trash
+     * @return void
+     */
+    public function index()
+    {
+        $tableList = [];
+        $tables = ConnectionManager::get('default')->schemaCollection()->listTables();
+        foreach ($tables as $tableName) {
```
softius: @mitsios similar functionality is provided by `App\SystemInfo\Database` . Suggest to use that to avoid having more than one implementations for the same thing.

softius: @mitsios we will need to consolidate the code between this implementation and `Database` class but let's do this in another PR once we are done with this feature

PR #707, opened 2018-11-22
CakePHP 3.6 deprecated error fixes (task #8244)

georgeconstantinou:
:information_source: All failing tests will be resolved in the feature branch **cakephp-36-task-8244**.

### DIFF
--- a/tests/TestCase/Controller/Api/ControllerApiTest.php
+++ b/tests/TestCase/Controller/Api/ControllerApiTest.php
@@ -80,10 +80,8 @@ class ControllerApiTest extends JsonIntegrationTestCase
     {
         $table = TableRegistry::getTableLocator()->get($module);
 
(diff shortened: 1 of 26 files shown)


tests/TestCase/Controller/Api/ControllerApiTest.php
```diff
@@ -169,7 +167,8 @@ public function modulesProvider(): array
     private function isModule(string $name): bool
     {
         $config = (new ModuleConfig(ConfigType::MIGRATION(), $name, '', ['cacheSkip' => true]))->parse();
-        $config = json_decode(json_encode($config), true);
+        $config = json_encode($config);
```
softius: @georgeconstantinou I think you can use `->parseToArray` instead to avoid the extra json checks

PR #708, opened 2018-11-22
PHPStan errors resolved (task #7909)

anvyst:


src/Event/Controller/Api/LookupActionListener.php
```diff
@@ -40,8 +40,13 @@ public function implementedEvents()
...
     {
-        $request = $event->getSubject()->request;
-        $table = $event->getSubject()->{$event->getSubject()->getName()};
+        /**
+         * @var \Cake\Controller\Controller $controller
+         */
+        $controller = $event->getSubject();
+        $request = $controller->getRequest();
+
+        $table = $controller->{$controller->getName()};
```
softius: @anvyst I believe you can replace this with `$controller->loadModel()`

PR #716, opened 2018-11-23
Trash controller (task #2555)

eliamariutti:
- `restore()` $table and $id to restore 
- `delete()`  $table and $id to permanently delete 
- `getMenuItems()` will provide a different menu for trashed records


src/Controller/TrashController.php
```diff
@@ -34,4 +34,72 @@ public function index()
...
+
+    /**
+     * Restore record
+     *
+     * @return void
+     * @throws \Cake\Network\Exception\NotFoundException When record not found.
+     */
+    public function restore()
+    {
+        $this->request->allowMethod(['get']);
```
mamchenkov: @eliamariutti I don't think GET is the appropriate method in these actions ..

PR #724, opened 2018-11-27
Fix missing type-hints for phpstan errors (task #7909)

anvyst:


src/Console/Installer.php
```diff
@@ -96,7 +96,7 @@ public static function postInstall(Event $event): void
      * @param \Composer\IO\IOInterface $io IO interface to write to console.
      * @return void
      */
-    public static function createAppConfig(string $dir, $io): void
+    public static function createAppConfig(string $dir, \Composer\IO\IOInterface $io): void
```
mamchenkov: @anvyst can we have the the class use statements at the top of the file instead of the namespaced ones in type hints please ..

anvyst: @mamchenkov fixed

PR #758, opened 2019-02-14
Moved API token expiration time into config (task #10034)

stepanov:


mamchenkov:
@stepanov we also have it in other places .. I see it in `src/Controller/Api/V1/V0/UsersController.php`

mamchenkov:
@stepanov there are some failing tests

stepanov:
> @stepanov there are some failing tests

Yes, I know, cannot reproduce it locally.

stepanov:
@mamchenkov @georgeconstantinou  The test failed because of error: 

`[Firebase\JWT\ExpiredException] Expired token`

Do you have idea why it happens?

anvyst:
@stepanov It happened because of wrong constant call within `Configure::read()` call. There's no `Api` key, but there's `API`. Due to that `expireTime` was returning `null`, thus the JWT token didn't have expiration time. This unit test derives from `extenal` group, which can be run using `./vendor/bin/phpunit --group external` to test it on local

PR #761, opened 2019-02-19
Changed advance search result for logs (Task #5037)

mitsios:
Changed the result cell for logs advance search action to be the same with the logs index page


src/Controller/LogsController.php
```diff
@@ -62,4 +68,104 @@ public function gc()
...
+
+        if (!$searchTable->isSearchable($model) && !$this->Auth->user('is_admin')) {
+            throw new BadRequestException('You cannot search in ' . implode(' - ', pluginSplit($model)) . '.');
+        }
+
+        if ($this->request->is('post')) {
+            $searchData = $search->prepareData($this->request);
+            if ('' !== $id) {
+                $search->update($searchData, $id);
+            }
```
softius: @mitsios you can use `else` here instead of another if

src/Controller/LogsController.php
```diff
@@ -62,4 +68,104 @@ public function gc()
...
+            'fields',
+            'aggregator',
+            'display_columns',
+            'sort_by_field',
+            'sort_by_order',
+            'group_by'
+        ];
+
+        $newData = [];
+        foreach ($searchFieldNamesAllowed as $fieldName) {
```
softius: @mitsios i believe this can be replaced with `array_intersect_key($data, $searchFieldNamesAllowed)`
```suggestion
        $searchFieldNamesAllowed = [
            'criteria' => null,
            'fields' => null,
            'aggregator' => null,
            'display_columns' => null,
            'sort_by_field' => null,
            'sort_by_order' => null,
            'group_by' => null,
        ];
        
        return array_intersect_key($data, $searchFieldNamesAllowed);
```

src/View/Cell/ResultLogsCell.php
```diff
@@ -0,0 +1,398 @@
...
+use Cake\Routing\Router;
+use Cake\Utility\Inflector;
+use Cake\View\Cell;
+use Cake\View\View;
+use InvalidArgumentException;
+use Search\Utility;
+use Search\Utility\Options;
+use Search\Utility\Search;
+
+final class ResultLogsCell extends Cell
```
softius: @georgeconstantinou is the same cell defined in `cakephp-search` ?

georgeconstantinou: @softius it does look the same, except there is an additional property `responseData`, but I do not see it being used anywhere.

PR #762, opened 2019-02-20
Remove API.auth from the project (task #10112)

eliamariutti:
- Remove all the instance of `API.auth` configuration
- Remove `API_AUTHENTICATION` from dotenv

### DIFF
--- a/config/bootstrap.php
+++ b/config/bootstrap.php
@@ -323,6 +323,10 @@ Configure::load('admin_lte', 'default');
  */
 Configure::load('system_info', 'default');
 
+if (Configure::read('API.auth') === false) {
+    Log::write('critical', "Non-authenticated API requests are deprecated");
+}
+
 /*
  * Feature Factory initialization
  * IMPORTANT: this line should be placed at the end of the bootstrap file.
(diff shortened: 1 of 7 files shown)


config/bootstrap.php
```diff
@@ -249,17 +249,12 @@
...
-if (Configure::read('Swagger.crawl') && Configure::read('API.auth')) {
+if (Configure::read('Swagger.crawl')) {
     Plugin::load('Alt3/Swagger', ['bootstrap' => true, 'routes' => true]);
 }
 Plugin::load('AdminLTE', ['bootstrap' => true, 'routes' => true]);
 
-/*
- * Only load JwtAuth plugin if API authentication is enabled
- */
-if (Configure::read('API.auth')) {
```
softius: @eliamariutti @georgeconstantinou I suggest to log a warning in case the `API.auth` exists and it set to `false`.

config/bootstrap.php
```diff
@@ -328,6 +323,10 @@
  */
 Configure::load('system_info', 'default');
 
+if (Configure::read('API.auth') === false) {
```
georgeconstantinou: @eliamariutti I think here you need to check for the dotenv variable instead since the configuration you are checking is already dropped..

PR #763, opened 2019-02-21
Enforce api authentication (Task #10112)

eliamariutti:
Check `API_AUTHENTICATION` on dotevn  file instead of `Configure::read('API.Auth')`

### DIFF
--- a/config/bootstrap.php
+++ b/config/bootstrap.php
@@ -323,7 +323,7 @@ Configure::load('admin_lte', 'default');
  */
 Configure::load('system_info', 'default');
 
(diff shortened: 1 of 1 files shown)


config/bootstrap.php
```diff
@@ -323,7 +323,7 @@
  */
 Configure::load('system_info', 'default');
 
-if (Configure::read('API.auth') === false) {
+if ((bool)env('API_AUTHENTICATION') === false) {
```
georgeconstantinou: @eliamariutti I think you need to combine this if statement with a not null check. The log needs to trigger if the 'API_AUTHENTICATION' parameter **is** present and is `false`.

PR #764, opened 2019-02-22
Api schema ending point (Task #10113)

eliamariutti:
`/api/<moduleName>/schema` will return the request module DB schema


src/Controller/Api/V1/V0/AppController.php
```diff
@@ -135,6 +135,15 @@ protected function _authentication(): void
...
     }
 
+    /**
+     * Custum schema action : return the schema of the request module
+     * @return \Cake\Http\Response|void|null
+     */
+    public function schema()
+    {
+        return $this->Crud->execute();
+    }
```
georgeconstantinou: @eliamariutti this method is not required. But you have to register the new action in the top of this file (like related action in line 40).

eliamariutti: @georgeconstantinou even if I register the action, if I don't write this method is not working. I got the expample from the Crud documentation https://crud.readthedocs.io/en/latest/actions/custom.html#using-custom-named-controller-actions . Am I missign sth ?

georgeconstantinou: let me re-test this locally but for me, it was working without the AppController method.

src/Crud/Action/SchemaAction.php
```diff
@@ -0,0 +1,66 @@
...
+                'model' => $association->getTarget()->getTable(),
+                'type' => $association->type(),
+                'primary_key' => $association->getBindingKey(),
+                'foreign_key' => $association->getForeignKey()
+            ];
+        }
+
+        $subject = $this->_subject(['success' => true]);
+
+        $this->_controller()->set('data', ['fields' => $data_fields, 'association' => $data_association]);
```
georgeconstantinou: @eliamariutti please change the "association" key to "associations".

src/Crud/Action/SchemaAction.php
```diff
@@ -0,0 +1,66 @@
...
+    protected function _get() : void
+    {
+        $controller = $this->_controller()->getName();
+        $mc = new ModuleConfig(ConfigType::MIGRATION(), $controller);
+        $fields = (array)$mc->parse();
+        $db_fields_type = TableRegistry::getTableLocator()->get($controller)->getSchema()->typeMap();
+
+        $data_fields = [];
+        foreach ($fields as $field => $value) {
+            $value->{'db_type'} = preg_match('/^(money|metric)/', $value->type) ? 'integer': $db_fields_type[$value->name];
```
georgeconstantinou: @eliamariutti we need to handle the combined (money, metric) fields in a different way. We need to remove them from the list (as they are not actual fields) and instead include the relevant database fields. You may be able to fetch this information from csv-migrations (see [here](https://github.com/QoboLtd/cakephp-csv-migrations/blob/v35.0.0/src/FieldHandlers/Provider/CombinedFields/MoneyCombinedFields.php)).

For example a combined money field with name _fee_ will create two database fields, _fee_amount_, and _fee_currency_. So the response data, instead of including this:
```json
{
    "name": "fee",
    "type": "money(currencies)",
    "required": false,
    "non-searchable": false,
    "unique": false,
    "db_type": "integer"
},
```

They must include this:
```json
{
    "name": "fee_amount",
    "type": "decimal",
    "required": false,
    "non-searchable": false,
    "unique": false,
    "db_type": "decimal"
},
{
    "name": "fee",
    "type": "currency(currencies)",
    "required": false,
    "non-searchable": false,
    "unique": false,
    "db_type": "string"
}
```

src/Crud/Action/SchemaAction.php
```diff
@@ -0,0 +1,66 @@
...
+            $data_fields[] = $value;
+        }
+
+        $data_association = [];
+        foreach ($this->_table()->associations() as $association) {
+            $data_association[] = [
+                'name' => $association->getName(),
+                'model' => $association->getTarget()->getTable(),
```
georgeconstantinou: Elia, for model property is better to fetch value from the class name, see example [here](https://github.com/QoboLtd/project-template-cakephp/blob/v45.0.0/src/Event/Controller/Api/BaseActionListener.php#L284).

src/Crud/Action/SchemaAction.php
```diff
@@ -0,0 +1,196 @@
...
+
+        $data_fields = [];
+        foreach ($fields as $field) {
+            $csvField = new CsvField($field);
+            $data = [
+                'name' => $csvField->getName(),
+                'type' => $csvField->getType()
+            ];
+
+            !empty($fieldJson[$csvField->getName()]['label']) ? $data['label'] = $fieldJson[$csvField->getName()]['label'] : '';
```
asaliev: Won't this not set `$data['label']` if `$fieldJson[$csvField->getName()]['label']` is empty, or am I missing something?
Something along these lines:
```php
$data['label'] = !empty($fieldJson[$csvField->getName()]['label']) ? $fieldJson[$csvField->getName()]['label'] : '';
```

eliamariutti: @asaliev because I don't want to set $data['label'] if is empty, so when array is encoded in json I'll not have `'label' : ''`  like required, non_searchable and unique.

src/Crud/Action/SchemaAction.php
```diff
@@ -0,0 +1,196 @@
...
+     *
+     * @return void
+     */
+    protected function _get() : void
+    {
+        $data_fields = [];
+        $data_association = $this->getAssociations($this->_table()->associations()->getIterator());
+
+        $method = 'getFieldsDefinitions';
+        if (method_exists($this->_table(), $method) && is_callable([$this->_table(), $method])) {
```
softius: @eliamariutti you can change this to `$table implements HasFieldsInterface`

src/Crud/Action/SchemaAction.php
asaliev: @eliamariutti gotcha

src/Crud/Action/SchemaAction.php
```diff
@@ -0,0 +1,191 @@
...
+     */
+    private function getOptionList(array $data) : array
+    {
+        $result = [];
+        foreach ($data as $key => $value) {
+            if (!empty($value['children'])) {
+                $value['children'] = $this->getOptionList($value['children']);
+            }
+
+            if (empty($value['inactive'])) {
```
georgeconstantinou: @eliamariutti here the logic should be slightly different, we need to exclude the option if inactive is set to `true`, the `unset` part can be executed always.

PR #765, opened 2019-02-25
IN clause suppor (task #9659)

eliamariutti:
IN clause support in conditions for API "index" endpoint. This will allow passing multiple values for a single column, using the following query string:
`/api/leads.json?conditions[name][]=foo&conditions[name][]=bar`
Will be converted to the following where clause:
`$conditions = [
    'Leads.name IN' => ['foo', 'bar']
];`

### DIFF
--- a/tests/TestCase/Controller/Api/V1/V0/UsersControllerTest.php
+++ b/tests/TestCase/Controller/Api/V1/V0/UsersControllerTest.php
@@ -198,5 +198,16 @@ class UsersControllerTest extends IntegrationTestCase
         $entity = $this->Users->get($id);
 
         $this->assertEquals($data['first_name'], $entity->get('first_name'));
+
+        $this->setAuthHeaders('00000000-0000-0000-0000-000000000002');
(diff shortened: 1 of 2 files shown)


softius:
@eliamariutti it would be great if we could add some tests for this.

tests/TestCase/Controller/Api/V1/V0/UsersControllerTest.php
```diff
@@ -198,5 +198,16 @@ public function testEditByLookupField(): void
...
+        $this->setAuthHeaders('00000000-0000-0000-0000-000000000002');
+        $this->get('/api/users/index.json?conditions[username][]=foo&conditions[username][]=user-1');
+        $this->assertResponseOk();
+
+        $body = $this->_response->getBody();
+        $response = json_decode($body);
+
+        $entity = $this->Users->get($id);
+
+        $this->assertEquals($data['first_name'], $entity->get('first_name'));
```
georgeconstantinou: @eliamariutti I think is better to move this into a separate method, something like `testIndexWithArrayConditions`.

PR #772, opened 2019-03-13
[WIP] Fixing batch editing flag

anvyst:
- Batch is handled by Search plugin and not CsvMigrations
  after index became a part of search plugin functionality


src/Controller/UsersController.php
```diff
@@ -29,7 +28,6 @@ class UsersController extends AppController
     use SimpleCrudTrait {
         SimpleCrudTrait::delete as defaultDelete;
     }
-    use U2fTrait;
```
softius: @anvyst can we postpone this until the next major version?

anvyst: @softius yes, we can put WIP tag on this PR.

PR #793, opened 2019-05-09
CI performance improvement (task #11384)

georgeconstantinou:
Currently, during a project installation/update, we first execute the CakePHP migrations against the test database to catch any errors before we run them against the live database.

This PR introduces a flag which is used to skip running migrations against the test database on CI builds since each build is run against a new environment without any pre-existing database data.

Listed below are some benchmarks run locally, executed three times for each case. The difference will dramatically increase in child projects where the number of tables is much greater.

Without `--quick` flag:
- 18.177s
- 19.495s
- 18.864s

With `--quick` flag:
- 11.243s
- 10.720s
- 11.139s

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ install:
   - ./bin/composer install --no-interaction --no-progress --no-suggest
   - yarn install
   - mkdir -p build/test-coverage build/test-results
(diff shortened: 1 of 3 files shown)


softius:
@georgeconstantinou let's change the option name to something more specific like `--skip-test-db`.

georgeconstantinou:
> @georgeconstantinou let's change the option name to something more specific like `--skip-test-db`.

@softius done https://github.com/QoboLtd/project-template-cakephp/pull/793/commits/c5c1c76593880e291f3712d5b2cf418f87a84a15

PR #794, opened 2019-05-17
Lookup Behavior (task #11230)

eliamariutti:
Move `LookupListener` to `LookupBehavior`


src/Model/Behavior/LookupBehavior.php
```diff
@@ -0,0 +1,306 @@
...
+        if (is_null($relatedEntity)) {
+            return;
+        }
+
+        $primaryKey = $association->getTarget()->getPrimaryKey();
+        if (!is_string($primaryKey)) {
+            throw new UnsupportedPrimaryKeyException();
+        }
+
+        $data[$association->getForeignKey()] = $relatedEntity->get($primaryKey);
```
georgeconstantinou: @eliamariutti let's adjust this method to actually return the value, and if the value returned is not empty, then we can adjust the `$data`.

src/Model/Behavior/LookupBehavior.php
```diff
@@ -0,0 +1,306 @@
...
+     *
+     * @param \Cake\ORM\Association $association Table association
+     * @param mixed $value Foreign key value
+     * @return bool
+     */
+    private function isValidID(Association $association, $value): bool
+    {
+        /**
+         * @var \Cake\ORM\Table $table
+         */
```
georgeconstantinou: @eliamariutti let's remove the annotation and use Assert library.

src/Model/Behavior/LookupBehavior.php
```diff
@@ -0,0 +1,306 @@
...
+        $table = $association->getTarget();
+        /**
+         * @var string $primaryKey
+         */
```
georgeconstantinou: @eliamariutti same here, let's use Assert library.

src/Model/Behavior/LookupBehavior.php
```diff
@@ -0,0 +1,306 @@
...
+
+    /**
+     * {@inheritDoc}
+     */
+    public function initialize(array $config)
+    {
+        parent::initialize($config);
+
+        if (!empty($this->getConfig('lookupFields'))) {
+            $this->lookupFields = $this->getConfig('lookupFields');
```
georgeconstantinou: @eliamariutti we need to probably assert that 'lookupFields' is an array.

src/Model/Behavior/LookupBehavior.php
```diff
@@ -0,0 +1,306 @@
...
+use Qobo\Utils\ModuleConfig\ConfigType;
+use Qobo\Utils\ModuleConfig\ModuleConfig;
+use Webmozart\Assert\Assert;
+
+/**
+ * Lookup behavior
+ */
+class LookupBehavior extends Behavior
+{
+    protected $lookupFields = null;
```
georgeconstantinou: @eliamariutti probably should set the default to an empty array, to reflect the expected/supported value.

PR #797, opened 2019-06-05
Update user link under groups (task #11510)

acrias:
The change will fix the user link when viewing users under a specific group

### DIFF
--- a/src/Template/Plugin/Groups/Groups/view.ctp
+++ b/src/Template/Plugin/Groups/Groups/view.ctp
@@ -59,7 +59,6 @@
                                 <thead>
                                     <tr>
                                         <th><?= __('Username') ?></th>
-                                        <th><?= __('Email') ?></th>
                                         <th><?= __('First Name') ?></th>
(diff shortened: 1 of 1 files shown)


src/Template/Plugin/Groups/Groups/view.ctp
```diff
@@ -0,0 +1,91 @@
...
+                                    <?php foreach ($group->users as $users) : ?>
+                                    <tr>
+                                        <td><?= $this->Html->link($users->username, [
+                                            'plugin' => false,
+                                            'controller' => 'Users',
+                                            'action' => 'view',
+                                            $users->id
+                                        ]) ?>
+                                        </td>
+                                        <td><?= h($users->email) ?></td>
```
eliamariutti: @andraskende It seems that the `$users->email` is always empty.

acrias: It seems that the email attribute is missing from the GroupsController

acrias: Email column removed

PR #798, opened 2019-06-06
Adding Vue Notifications area for search (task #11144)

anvyst:
- Adding `vue-notification` plugin for generic search notification areas for usability reasons.


softius:
@anvyst was `RecurrenceInput.vue ` added on purpose?

anvyst:
@softius we can keep it for later use - it's been part of base, but it can be used in template-cakephp as well. There're no changes in this component from here onwards till the clients.

PR #802, opened 2019-06-19
Fixing system search (task #11679)

anvyst:
- Resolving the problem with system search for search vuejs component (store)

### DIFF
--- a/resources/src/store/modules/search/index.js
+++ b/resources/src/store/modules/search/index.js
@@ -219,6 +219,10 @@ export default {
         value.content.saved.criteria = {}
       }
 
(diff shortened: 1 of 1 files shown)


resources/src/store/modules/search/index.js
```diff
@@ -268,12 +272,6 @@ export default {
         }).then(response => {
           if (response.data.success === true) {
             dispatch('savedSearchesGet')
-
```
softius: @anvyst did we remove the notifications here on purpose?

anvyst: @softius nope, let me patch it now.

PR #807, opened 2019-07-05
Refactoring Search & JS project architecture (task #11872)

anvyst:
### General Change

This PR was aimed on refactoring current Search components functionality.

Major changes involved:
* Restructure Search component for better unit testing (using `jest|vue-jest`)
* Adjust `TableAjax` component to work with recently modified Search.
* Reduce the load of `data ()` and `props` params in components. Instead rely more on the constants in `utils/` directory
* Detach API calls in separate object for readability and compact presentation.
* Replaced build instructions from `yarn` to `npm` for better compatibility among OS'es.
* Added `package-lock.json` from `npm` for locking dependencies (to avoid accidental upgrades). 
* Added `.editorconfig` to reflect indentation of Vue code inside `resources/` directory.

### Unit Testing / Acceptance Tests

Initial implementation using `karma|chai` didn't work well according to the number of unit tests being written in the system.
Instead, I'm proposing to integrate with `jest` as default testing framework. With `vue-jest` and `@vue/test-utils` we simplify the whole unit test process, and remove most of redundant/obsolete code.

### Running Tests

Current setup support multiple ways of Component tests.

* `npm run buid:dev` runs `devServer` from Node to launch at port 8080, where you can add ui components isolated from the PHP core, and test its functionality. Compiled JS is stored in-memory. Re-compiles everytime you make a change.
* `npm run watch` runs watch via Node, allowing to store unminimised version of JS in `webroot/dist/` directory allowing to use JS inside PHP project.
* `npm run lint` - runs ESLint to validate the code
* `npm run test` - performs Jest unit tests for Vue components.

### Components Structure

* AJAX endpoints communications are moved into `ApiService/` directory for usability, as well as having common source of setting up the work with `axios`.
* If we work with complex components - they can be stored i its separate directories (like `Dashboards`) with corresponding mixins code, and child-components.


resources/src/components/ui/TableAjax.vue
```diff
@@ -331,7 +367,7 @@ export default {
     },
 
     batchDelete () {
-      if (this.data.group_by) {
+      if (this.data.group_by || this.withBatch) {
```
georgeconstantinou: @anvyst the condition here I think it should be `! this.withBatch`. Like above on `batchEdit` method.

PR #809, opened 2019-07-09
Extend LookupBehavior to include a custom finder method (Task #11691)

acrias:
You can now extend the LookupBehavior using the custom finder method findLookup(). Further from the defined lookupfields you can add additional fields in the where clause.

Sample using where:
`$user_search = $table->find('Lookup', ['value' => 'username'])->where(['first_name' => 'yourname']);`

Sample:
`$user_search = $table->find('Lookup', ['value' => 'username']);`


src/Model/Behavior/LookupBehavior.php
```diff
@@ -297,4 +297,53 @@ private function getRelatedEntity(Association $association, ArrayObject $data, a
...
+     *
+     * @param \Cake\ORM\Query $query Query object
+     * @param mixed[] $options Query options
+     * @return \Cake\ORM\Query|null
+     */
+    public function findLookup(Query $query, array $options): ?Query
+    {
+        $table = $this->_table;
+
+        $tableAlias = $this->_table->getAlias();
```
georgeconstantinou: @acrias I do not see this variable being used anywhere.

src/Model/Behavior/LookupBehavior.php
```diff
@@ -297,4 +297,53 @@ private function getRelatedEntity(Association $association, ArrayObject $data, a
...
+            $primaryKey = $table->getPrimaryKey();
+            if (! is_string($primaryKey)) {
+                throw new UnsupportedPrimaryKeyException();
+            }
+
+            $query->where([
+                $table->aliasField($primaryKey) => $options['value']
+            ]);
+
+            return null;
```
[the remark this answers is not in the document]
acrias: There is a mistake here, it should returned the constructed query instead of null. So either it returns the query without the lookup fields into consideration or it returns the query with the lookup fields

src/Model/Behavior/LookupBehavior.php
```diff
@@ -309,14 +309,18 @@ public function findLookup(Query $query, array $options): ?Query
...
         if (! isset($options['value'])) {
-            return null;
+            return $query;
         }
 
+        /**
+         * In case lookupfields in table config are not empty and we want to use the custom finder
+         * we can do it by sending in the options an empty array in lookupfields
+         */
+        $ignoreLookupFields = (isset($options['lookupfields']) && !$options['lookupfields']) ? true : false;
```
eliamariutti: @acrias Does this line do the some job of `empty()` ?

acrias: As far as I know if the index 'lookupfields' does not exist generates a warning.

georgeconstantinou: @acrias I am not sure what is the usage of this. If we want to bypass the lookup fields logic is better to simply call the default finder `$table->find('all')`, instead of calling the _lookup_ finder.

tests/TestCase/Model/Behavior/LookupBehaviorTest.php
```diff
@@ -125,4 +125,23 @@ public function testfindLookupWithWhere() : void
         $email = is_array($query) ?: $query->get('email');
         $this->assertSame('user-1@test.com', $email);
     }
+
+    public function testfindLookupIgnoreLookupFieldsInConfig() : void
+    {
+        $query = $this->users->find('lookup', ['value' => '00000000-0000-0000-0000-000000000002', 'lookupfields' => false])->firstOrFail();
+        $email = is_array($query) ?: $query->get('email');
```
georgeconstantinou: @acrias if the result is an array you will need to retrieve the email like this: `$email = $query['email']`.

src/Model/Behavior/LookupBehavior.php
acrias: You can do this yes but you will loose the default primary key search. You will have to use the where in the `$table->find('all')->where(['id' => '86687687'])`. Instead you can use the `$table->find('lookup', ['value' => '00000000-0000-0000-0000-000000000002', 'lookupfields' => false])`. It depends where we want to call the lookup finder and in what cases

PR #811, opened 2019-07-12
Fix Dynamic panel edit batch (task #10052)

eliamariutti:
The method  `filterFields()` was removing incorrectly all the fields of hidden panels.


src/Controller/Component/CsvViewComponent.php
```diff
@@ -87,15 +87,6 @@ protected function filterFields(Event $event) : void
...
 
-        $panels = $this->getPanels($config, $controller->viewVars['entity']->toArray());
-        if (! empty($panels['fail'])) {
-            // filter out fields of hidden panels
-            $controller->viewVars['fields'] = array_diff_key(
-                $controller->viewVars['fields'],
-                array_flip($panels['fail'])
-            );
-        }
-
```
georgeconstantinou: @eliamariutti if we remove this code, all panels are visible in record's detailed view, which is not the expected behavior.

georgeconstantinou: We will probably need to skip this logic only if is a batch action.

PR #815, opened 2019-07-18
Log actions request on controllers (task #11957)

eliamariutti:
We can choose which controller and action will be logged in the AuditStash table or to file.


config/log_actions.php
```diff
@@ -0,0 +1,11 @@
+<?php
+return [
+    'LogActions' => [
+        // List of controllers to Log
+        'controllers' => ['Things'],
+        // Actions of the controllers to log
+        'actions' => ['view', 'edit'],
```
anvyst: @eliamariutti @georgeconstantinou won't it be easier to replace `actions` array of all the actions you want to keep records upon, into `excludeActions` - and technically, track everything for this controller excluding the ones listed in this property? There can be a lot of actions, that people will forget to pre-populate in this configuration.

src/Controller/Component/LogActionsComponent.php
```diff
@@ -0,0 +1,90 @@
...
+        $table->save(new Entity($data));
+    }
+
+    /**
+     * Log to file
+     * https://en.wikipedia.org/wiki/Common_Log_Format
+     *
+     * @return void
+     */
+    protected function logToFile() : void
```
anvyst: @softius @eliamariutti @georgeconstantinou we already have `monolog/monlog` as part of project setup. 
Won't it be easier to hand over the creation of the files and storing this information using monolog, instead of handling the files on our own?

softius: I agree. AuditStash also provides PersisterInterface, just have a look at `App\Persister\MysqlPersister` on this project.

src/Controller/Component/LogActionsComponent.php
```diff
@@ -0,0 +1,90 @@
...
+
+        $data = [
+            'timestamp' => Time::parse('now'),
+            'primary_key' => empty($this->request->getParam('pass')[0]) ? '' : $this->request->getParam('pass')[0],
+            'source' => $this->request->getParam('controller'),
+            'user_id' => $user_id,
+            'meta' => json_encode($meta)
+        ];
+
+        $table->save(new Entity($data));
```
softius: @eliamariutti @georgeconstantinou I think it would be better to use the same mechanism AuditStash does, that is by throwing an event instead of adding the record direclty.

config/Migrations/20190722063329_AddActionLogAudit.php
```diff
@@ -0,0 +1,23 @@
...
+     * Change Method.
+     *
+     * More information on this method is available here:
+     * http://docs.phinx.org/en/latest/migrations.html#the-change-method
+     * @return void
+     */
+    public function change()
+    {
+        $table = $this->table('log_audit');
+        $table->addColumn('action', 'string', [
```
georgeconstantinou: No need to introduce this column. We can keep this info under the _meta_ column. Instead, we need to introduce columns for existing fields _type_ and _transaction_, which are missing from our table.

src/Persister/MysqlPersister.php
```diff
@@ -19,6 +19,7 @@ public function logEvents(array $auditLogs)
             $eventType = $log->getEventType();
             $meta = $log->getMetaInfo();
             $data = [
+                'action' => !empty($meta['action']) ? $meta['action'] : null,
```
georgeconstantinou: @eliamariutti let's move this info in `$meta`.

config/log_actions.php
georgeconstantinou: @anvyst we need to be very specific on which actions are considered as `reading` the actual record data.

src/Event/AuditViewEvent.php
```diff
@@ -0,0 +1,17 @@
...
+class AuditViewEvent extends BaseEvent
+{
+    /**
+     * Returns the type name of this event object.
+     *
+     * @return string
+     */
+    public function getEventType()
+    {
+        return 'view';
```
georgeconstantinou: @eliamariutti to keep it consistent with the existing event types, this should be _read_.

PR #827, opened 2019-08-05
 Personal data clear (Task #11954)

eliamariutti:
If a module has personal data marked in the 'fields.json', in the view will show a new "Clear personal data" button.
Once clicked, from the record will be erased all the fields marked as personal after a configured period.
The shell script will take cake of pick all the expired records of a selected module (or all the them), and erase the data.


config/Migrations/20190801083855_ScheduledPersonalData.php
```diff
@@ -0,0 +1,59 @@
...
+    /**
+     * Change Method.
+     *
+     * More information on this method is available here:
+     * http://docs.phinx.org/en/latest/migrations.html#the-change-method
+     * @return void
+     */
+    public function change()
+    {
+        $table = $this->table('scheduled_personal_data', ['id' => false, 'primary_key' => ['id']]);
```
acrias: Is this needed as we have below the addPrimaryKey?

eliamariutti: @acrias Actually this is needed due phinx add an incremental id field (https://book.cakephp.org/3.0/en/phinx/migrations.html#creating-a-table), but I will drop the `$table->addPrimaryKey(['id']);`

src/Shell/ClearPersonalDataShell.php
```diff
@@ -0,0 +1,116 @@
...
+     * @return mixed[]
+     */
+    private function getPersonalField(string $moduleName) : array
+    {
+        $config = (array)(new ModuleConfig(ConfigType::FIELDS(), $moduleName))->parseToArray();
+        $fields = array_filter($config, function ($v) {
+            return in_array('personal', (array)array_keys($v));
+        });
+
+        return array_map(function () {
```
georgeconstantinou: @eliamariutti is this returning key-value pairs, with personal fields as keys and empty strings as value?

eliamariutti: yes, exactly!

PR #829, opened 2019-08-13
Add Traits folder (Task #12955)

eliamariutti:
Move all Traits in a separate folder


config/bootstrap.php
```diff
@@ -85,7 +85,7 @@
     Configure::load('icons', 'default');
     Configure::load('menu', 'default');
     Configure::load('roles_capabilities', 'default');
-    Configure::load('scheduled_log', 'default');
+    Configure::load('scheduled_task', 'default');
```
softius: @eliamariutti is this necessary?

eliamariutti: @softius the idea is to store all the value for scheduled task in one single file

PR #849, opened 2019-09-19
Chicken egg issue with settings (task #13364)

eliamariutti:
The cache were causing issues with the settings table after removing and reinstalling a project

### DIFF
--- a/build/Robo/Command/App/App.php
+++ b/build/Robo/Command/App/App.php
@@ -2,6 +2,7 @@
 
 namespace Qobo\Robo\Command\App;
 
+use Cake\Shell\SchemaCacheShell;
 use Qobo\Robo\AbstractCommand;
 use Qobo\Robo\Runner;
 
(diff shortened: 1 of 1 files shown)


build/Robo/Command/App/App.php
```diff
@@ -81,6 +81,8 @@ public function appUpdate($env = '', array $options = ['skip-test-db' => false])
      */
     public function appRemove()
     {
+        // clear ORM cache
+        $this->taskCakephpShellScript()->name('orm_cache')->param('clear')->run();
```
softius: @eliamariutti perhaps we should do `./bin/cake cache clear_all`, for the same reason?

eliamariutti: @softius the problem is the ORM cache, the file `tmp/cache/models/myapp_cake_model_default_settings` . Cleaning the "normal" cache not sure is useful.

eliamariutti: Re-thinking about this, it is a good idea to have a clean project folder when we remove it.

PR #854, opened 2019-10-04
Search UI aggregates (task #12103)

georgeconstantinou:
This PR introduces aggregates on search UI along with the following search-related codebase improvements/cleanup:
- Split search component into multiple sub-components
- Adjusted existing saved searches to the new database table structure (migration)
- Set count aggregate for existing saved searches that use group by clause
- Set window URL to currently saved search ID
- Simplified vuex search module state structure
- Simplified batch detection and execution logic
- Moved reset logic within the store
- Adjusted search-related Views to properly pass aggregate field headers to Vue ajax table
- Simplified search manager codebase, filters mapping logic has been moved into the plugin
- Adjusted charts to work with aggregates

This PR depends on the following PRs:
- https://github.com/QoboLtd/cakephp-search/pull/262
- https://github.com/QoboLtd/cakephp-search/pull/264
- https://github.com/QoboLtd/cakephp-search/pull/265


resources/src/components/Search/AggregateSelector.vue
```diff
@@ -0,0 +1,150 @@
...
+  watch: {
+    criteria(value) {
+      const aggregate = Object.keys(value).find(item => Aggregate.isAggregate(item))
+      // if criteria include an aggregate, populate guid, filter and value inputs from it
+      if (aggregate !== undefined) {
+        const guid = Object.keys(value[aggregate])[0]
+        this.guid = guid
+        this.filter = value[aggregate][guid].operator
+        this.value = value[aggregate][guid].value
+      } else {
```
anvyst: @georgeconstantinou is there a special need for `else` statement, our you can simply pre-define these variables before the `if()`?

resources/src/components/Search/AggregateSelector.vue
```diff
@@ -0,0 +1,150 @@
...
+        this.filter = ''
+      }
+    },
+    fields(value) {
+      // if fields include an aggregate, populate aggregate and field inputs from it
+      if (Aggregate.hasAggregate(value)) {
+        this.aggregate = Aggregate.extractAggregateType(Aggregate.getAggregate(value))
+        this.field = Aggregate.extractAggregateField(Aggregate.getAggregate(value))
+      } else {
+        this.aggregate = ''
```
anvyst: Same here, for the `else` statement

resources/src/components/Search/AggregateSelector.vue
```diff
@@ -0,0 +1,150 @@
...
+        this.field = ''
+      }
+    }
+  },
+  created() {
+    this.$watch(vm => [vm.aggregate, vm.field, vm.filter, vm.model, vm.value], (newValue, oldValue) => {
+      this.$store.commit('search/criteriaRemove', this.guid)
+      if (this.aggregate && this.field && this.filter && this.value) {
```
anvyst: This would look nice as a computed variable/method instead of a long complex `if()`

resources/src/components/Search/AggregateSelector.vue
georgeconstantinou: Unfortunately yes, since these values are being watched (line 91), so pre-defining them will trigger the watcher.

resources/src/components/Search/AggregateSelector.vue
georgeconstantinou: Same here, these properties are watched.

PR #859, opened 2019-10-14
Add cache to Settings (task #13363)

eliamariutti:
Add cache to Settings

### DIFF
--- a/src/Settings/DbConfig.php
+++ b/src/Settings/DbConfig.php
@@ -1,6 +1,7 @@
 <?php
 namespace App\Settings;
 
+use Cake\Cache\Cache;
 use Cake\Core\Configure\ConfigEngineInterface;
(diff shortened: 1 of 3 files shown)


src/Settings/DbConfig.php
```diff
@@ -31,15 +32,21 @@ public function read($key)
...
+            try {
+                $data = $query->find('list', ['keyField' => 'key', 'valueField' => 'value'])
+                              ->where(['scope' => $this->scope, 'context' => $this->context])
+                              ->toArray();
+            } catch (\Cake\Database\Exception $e) {
+                return [];
+            }
+
+            $config = Hash::expand($data);
+            Cache::write('Settings', $config);
```
softius: @eliamariutti we need also to invalidate the cache, whenever that is necessary.

PR #863, opened 2019-10-23
Initial version for fixing datetime fields (Task #4336)

acrias:
This PR will try to update all records from specific modules having their field type 'datetime'. It will convert the date from one timezone to another.

Sample command:
`./bin/cake fix_date_time --module=Things --timezonefrom="Asia/Nicosia" --timezoneto="UTC"`


src/Shell/FixDateTimeShell.php
```diff
@@ -0,0 +1,176 @@
...
+    public function main(string $modules = '')
+    {
+        $this->info('Checking for datetime fields to be updated');
+        $this->hr();
+
+        $this->modules = !empty($this->param('module')) ? (array)$this->param('module') : Utility::getModules();
+
+        if (empty($this->modules)) {
+            $this->warn('Did not find any modules');
+            exit();
```
georgeconstantinou: @acrias you can return `false` here instead of exiting, as this will break the execution.

src/Shell/FixDateTimeShell.php
```diff
@@ -0,0 +1,176 @@
...
+     *
+     * @param string $module Module name
+     * @return int Number of encountered errors
+     */
+    public function checkFields(string $module) : int
+    {
+        $mc = $this->getModuleConfig($module, []);
+
+        $fields = [];
+        $config = json_encode($mc->parse());
```
georgeconstantinou: @acrias you can use `parseToArray` method instead, to avoid the next line.

src/Shell/FixDateTimeShell.php
```diff
@@ -0,0 +1,176 @@
...
+                    continue;
+                }
+                $message = 'Record [' . $entity->get('id') . ']. Value of ' . $fieldToUpdate . ' changed from ' . $entity->get($fieldToUpdate);
+
+                $datetime = new \Cake\I18n\Time($entity->get($fieldToUpdate)->format('Y-m-d H:i:s'), $this->params['timezonefrom']);
+                $datetime = $datetime->setTimezone($this->params['timezoneto']);
+                $entity->set($fieldToUpdate, $datetime);
+                $table->saveOrFail($entity);
+                $message .= ' to ' . $entity->get($fieldToUpdate);
+                $this->warn($message);
```
georgeconstantinou: @acrias lets use `info` instead of `warn` here.

src/Shell/FixDateTimeShell.php
```diff
@@ -0,0 +1,176 @@
...
+     */
+    public function checkDateTimeFields(string $module, array $fields = []): void
+    {
+        $skipfields = ['created', 'modified', 'trashed'];
+        $fieldsToUpdate = [];
+
+        $this->info('Trying to update datetime fields for ' . $module);
+
+        $table = TableRegistry::getTableLocator()->get($module);
+        $entities = $table->find();
```
georgeconstantinou: @acrias this will probably run PHP out of memory for large tables. One improvement is to use `select` method and provide only the `$fields` and the primary key (required later on for the update operation, but do not hardcode `id` here, instead use `$table->getPrimaryKey()` method).

In general though, I think we need to paginate through the table records, instead of fetching all of them.

src/Shell/FixDateTimeShell.php
acrias: @georgeconstantinou I think also we have to consider the option to marked which records have been updated so in case of something goes wrong and the shell stop, to know which have been updated. Also we cannot execute the shell more than once since it will update the datetime again.

georgeconstantinou: Correct, we will probably need to switch this into a CakePHP migration instead of a standalone shell script. Migrations are executed once.

src/Shell/FixDateTimeShell.php
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    /**
+     * Check fields and their types.
+     *
+     * @param string $module Module name.
+     * @param mixed[] $fields List of field definitions.
+     * @return void
+     */
+    public function updateDateTimeFields(string $module, array $fields = []): void
```
softius: @acrias this is a long method, with high complexity and no tests. Please lets try to break this into small pieces with the necessary tests in place.

PR #879, opened 2019-11-22
Fix dynamic panel issue (Task #14073)

acrias:
This PR fixes the dynamic panels issue that were remaining hidden in case the `file` fields showing in edit/add json were at the beginning

### DIFF
--- a/src/Controller/Api/V1/V0/AppController.php
+++ b/src/Controller/Api/V1/V0/AppController.php
@@ -482,9 +482,7 @@ class AppController extends Controller
         }
 
         if (is_array($data[$this->name])) {
-            $innerKey = key($data[$this->name]);
-            // embedded form - [module][dynamicField][inputName] : Regular form format - [module][inputName]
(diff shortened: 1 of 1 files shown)


src/Controller/Api/V1/V0/AppController.php
```diff
@@ -482,9 +482,7 @@ public function panels()
         }
 
         if (is_array($data[$this->name])) {
-            $innerKey = key($data[$this->name]);
-            // embedded form - [module][dynamicField][inputName] : Regular form format - [module][inputName]
-            $data = is_array($data[$this->name][$innerKey]) ? $data[$this->name][$innerKey] : $data[$this->name];
+            $data = $data[$this->name];
```
softius: @georgeconstantinou @acrias it looks like we are removing some functionality here - perhaps embedded fields within dynamic panels?

georgeconstantinou: It works fine for me. I think it might be a very old leftover when embedded module fields were part of the same form as the current module, not 100% sure though.

PR #889, opened 2019-12-05
Bugfix for display columns label (task #14433)

georgeconstantinou:
This PR fixes the issue of incorrect label for the display column if the field is the first one in the available search filters.


georgeconstantinou:
> @georgeconstantinou can we try and have a test for this?

@softius test has been added https://github.com/QoboLtd/project-template-cakephp/pull/889/commits/9c74b4710bff153870ef8650417261575ba50056

PR #893, opened 2019-12-06
Translations for project template cakephp (Task #14435)

acrias:
This PR updates translation and added localization files


softius:
We are leaving app-level messages under the default domain.

PR #901, opened 2019-12-18
Internationalization and Localization (Task #13980)

acrias:


acrias:
Initial commit for the translations

config/cron.php
```diff
@@ -9,6 +9,7 @@
                 'PluginShell',
                 'CronShell',
                 'FixDateTimeShell',
+                'GenerateLanguageFilesShell',
```
softius: @acrias I don't think that we need this as part of the cron job. The command must be executed only before updating the pot file.

src/Locale/el_GR/default.po
```diff
@@ -0,0 +1,1363 @@
+# LANGUAGE translation of CakePHP Application
```
softius: @acrias We can drop this file since we are not adding any actual translation.

src/Shell/GenerateLanguageFilesShell.php
```diff
@@ -0,0 +1,272 @@
...
+use CsvMigrations\FieldHandlers\FieldHandlerFactory;
+use CsvMigrations\Utility as CsvUtility;
+use CsvMigrations\Utility\Panel;
+use CsvMigrations\Utility\Validate\Utility;
+use PHPUnit\Framework\Assert;
+use Qobo\Utils\ModuleConfig\ConfigType;
+use Qobo\Utils\ModuleConfig\ModuleConfig;
+use Qobo\Utils\ModuleConfig\Parser\Parser;
+
+class GenerateLanguageFilesShell extends BaseShell
```
softius: @acrias it might be better to create a bake shell, rather a regular shell. This way you won't have to deal with generating the template or updating the files.

src/Shell/GenerateLanguageFilesShell.php
```diff
@@ -0,0 +1,272 @@
...
+        //Module Alias
+        $ctpLines .= "//Module: " . $module . ", Module Alias: " . $table->getAlias() . "\n";
+        $ctpLines .= $this->generateCtpLine($table->getAlias()) . "\n\n";
+
+        //Field Labels
+        $fieldLabelConfig = (new ModuleConfig(ConfigType::FIELDS(), $module))->parseToArray();
+        foreach ($fieldLabelConfig as $key => $fieldLabel) {
+            if (isset($fieldLabel['label'])) {
+                $ctpLines .= "//Module: " . $module . ", Field Label for: " . $key . "\n";
+                //var_dump($fieldLabel['label']);
```
softius: @acrias remove any dead / commented out code.

src/Template/Module/Translations/Common.ctp
```diff
@@ -0,0 +1,35 @@
+<?php
+
+//Module: Common, Module Title: Common
```
softius: @acrias is there any way that we could display the actual JSON file?

acrias: In some cases we have found a way as per @georgeconstantinou  suggestion but we have some more files to find for example the view files (edit, add, view, index)

src/Shell/GenerateLanguageFilesShell.php
```diff
@@ -0,0 +1,326 @@
...
+
+        $mc = $this->getModuleConfig($module, []);
+        $configFile = (string)$mc->find(false);
+
+        //Module Title and Alias
+        $ctpLines .= $this->generateCommentCtpLine(str_replace(CONFIG, 'config' . DS, $configFile)) . "\n";
+        $ctpLines .= $this->generateCommentCtpLine("Module Title") . "\n";
+        $ctpLines .= $this->generateCtpLine(Inflector::humanize(Inflector::underscore($module))) . "\n";
+        $ctpLines .= $this->generateCommentCtpLine("Module Alias") . "\n";
+        $ctpLines .= $this->generateCtpLine($table->getAlias()) . "\n\n";
```
softius: @acrias this must come from the `config.dist.json` or `config.json` of each module. There is an optional property `table.alias` that can be used to overwrite module title.

src/Shell/GenerateLanguageFilesShell.php
```diff
@@ -0,0 +1,326 @@
...
+     * Generate file for each module.
+     * @param string $module Module name
+     * @return void
+     */
+    public function generateCtpFiles(string $module): void
+    {
+        $factory = new FieldHandlerFactory();
+        $table = TableRegistry::getTableLocator()->get($module);
+
+        $callable = [$this, 'generateCtpLine'];
```
softius: @acrias I believe this block can be completely removed. It looks like a leftover to me.

src/Shell/GenerateLanguageFilesShell.php
```diff
@@ -0,0 +1,326 @@
...
+            $this->generateCtpFiles((string)$module);
+        }
+    }
+
+    /**
+     * Generate file for each module.
+     * @param string $module Module name
+     * @return void
+     */
+    public function generateCtpFiles(string $module): void
```
softius: @acrias as a general comment, I would suggest dropping the `Ctp` part from these methods. So we are going to have `generateFiles` instead of `generateCtpFiles`, `generateCommentLine` instead of `generateCtpCommentLine`

src/Shell/GenerateLanguageFilesShell.php
```diff
@@ -0,0 +1,326 @@
...
+                sprintf("Method %s.%s does not exist", get_class($callable[0]), $callable[1])
+            );
+        }
+
+        $ctpLines = "<?php\n\n";
+
+        $mc = $this->getModuleConfig($module, []);
+        $configFile = (string)$mc->find(false);
+
+        //Module Title and Alias
```
softius: @acrias please try to move each code block to a separated method (protected or private) to improve code readability. By code block I refer to the lines between each commented section

PR #902, opened 2019-12-20
Raw formatter on rest API (task #14452)

georgeconstantinou:
This PR adjusts the rest API `GET` methods to rely on the new RawFormatter class instead of using custom methods for handling response formatting.


tests/TestCase/Controller/Api/V1/V0/ThingsControllerTest.php
```diff
@@ -149,8 +149,8 @@ public function rawDataProvider(): array
                     'testmetric_unit' => 'ft',
                     'testmoney_amount' => 155.22,
                     'testmoney_currency' => 'USD',
-                    'created' => '2018-01-18T15:47:16+00:00',
-                    'modified' => '2018-01-18T15:47:16+00:00',
+                    'created' => '2018-01-18 15:47:16',
```
softius: @georgeconstantinou we are dropping important information here which is the timezone. We also switch from the ISO format to something else. Any way we can keep the old format?

georgeconstantinou: @softius I adjusted the _RawFormatter_ class to use the ISO format: https://github.com/QoboLtd/project-template-cakephp/pull/902/commits/853c9d4390ccb25ee351e77f13966dbde1dcf9b7

PR #908, opened 2020-01-07
Update username validator (Task #14656)

acrias:
This PR removes the space character from the allow character list

### DIFF
--- a/src/Model/Table/UsersTable.php
+++ b/src/Model/Table/UsersTable.php
@@ -71,20 +71,20 @@ class UsersTable extends Table
         $validator = parent::validationDefault($validator);
 
         $validator->add('username', 'validRegex', [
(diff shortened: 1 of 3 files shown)


src/Model/Table/UsersTable.php
```diff
@@ -71,20 +71,20 @@ public function validationDefault(Validator $validator)
         $validator = parent::validationDefault($validator);
 
         $validator->add('username', 'validRegex', [
-            'rule' => ['custom', '/^[\w\d\@\-\_\s\.]+$/Du'],
-            'message' => 'The provided value is invalid (alphanumeric, dot, dash, at, underscore, space)',
+            'rule' => ['custom', '/^[\w\d\_\.\@\-]+$/Du'],
+            'message' => __('The provided value is invalid (alphanumeric, underscore, dash, at)'),
```
georgeconstantinou: @acrias let's add back the _dot_ word in the error message.

PR #914, opened 2020-01-16
adding Active class for currently viewed URL (task #14956)

anvyst:
* Adding `active` class for current URL the user is on.

### DIFF
--- a/webroot/js/general.js
+++ b/webroot/js/general.js
@@ -3,8 +3,8 @@
 
     //Highlight active link in the sidebar.
     var url = window.location;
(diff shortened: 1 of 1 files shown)


webroot/js/general.js
```diff
@@ -1,6 +1,12 @@
 ;(function ($) {
     'use strict';
 
+    //Highlight active link in the sidebar.
+    var url = window.location;
+    $('.sidebar ul.sidebar-menu a').filter(function() {
+        return this.href == url;
```
asaliev: We need to trim the trailing slash from `this.href` and `url` before comparing them

PR #925, opened 2020-02-10
Missing template error fix (task #5611)

georgeconstantinou:
This PR fixes the error shown below by rendering the default layout, instead of a custom error layout, when error templates are loaded.

This means that post-login and if there are any errors, they are displayed on the current page instead of being redirected to `/system/error`. Pre-login, instead of an error page the visitor is redirected to the login page.

**Fixed error:**
```php
Fatal Error (256): [Cake\View\Exception\MissingTemplateException] Template file "System/json/error.ctp" is missing.
```


src/Controller/ErrorController.php
```diff
@@ -43,6 +43,7 @@ public function initialize()
      */
     public function beforeFilter(Event $event)
     {
+        parent::beforeFilter($event);
```
eliamariutti: @georgeconstantinou what is the difference between this and delete all the method ? some for `afterFilter`

georgeconstantinou: no difference, I will remove them now.

PR #942, opened 2020-03-06
Filters caching (task #15571)

georgeconstantinou:
Skip caching if search filters result is empty.


softius:
@georgeconstantinou can you please add a small comment explaining why the change happened?

PR #954, opened 2020-03-24
Lookup endpoint in swagger documentation (Task #15805)

acrias:
This PR adds the lookup endpoint in the swagger documentation.

### DIFF
--- a/src/Swagger/Annotation.php
+++ b/src/Swagger/Annotation.php
@@ -154,7 +154,7 @@ class Annotation
             )
 
             @SWG\Get(
-                path="/api/{{module_url}}/lookup.json",
+                path="/api/{{module_url}}/lookup",
                 summary="Retrieve a list of records {{module_human_singular}} by name",
                 tags={"{{module_human_plural}}"},
                 produces={"application/json"},
--- a/src/Swagger/Annotation.php
+++ b/src/Swagger/Annotation.php
@@ -153,6 +153,90 @@ class Annotation
                 )
             )
 
+            @SWG\Get(
+                path="/api/{{module_url}}/lookup.json",
+                summary="Retrieve a list of records {{module_human_singular}} by name",
+                tags={"{{module_human_plural}}"},
+                produces={"application/json"},
+                @SWG\Parameter(
+                    name="query",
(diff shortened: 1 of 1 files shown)


src/Swagger/Annotation.php
```diff
@@ -153,6 +153,90 @@ class Annotation
                 )
             )
 
+            @SWG\Get(
+                path="/api/{{module_url}}/lookup.json",
```
georgeconstantinou: @acrias I think we can remove the .json extension here to keep it consistent with the rest of the endpoints.

src/Swagger/Annotation.php
```diff
@@ -153,6 +153,90 @@ class Annotation
...
+                summary="Retrieve a list of records {{module_human_singular}} by name",
+                tags={"{{module_human_plural}}"},
+                produces={"application/json"},
```
georgeconstantinou: @acrias we can also include the `consumes` parameter:
```
consumes={"application/json"}
```

src/Swagger/Annotation.php
acrias: I think I have tried it too but it does not work

src/Swagger/Annotation.php
acrias: @georgeconstantinou is it needed? because we are not sending json is get and the parameters is on the url

georgeconstantinou: true, we can skip this.

src/Swagger/Annotation.php
acrias: @georgeconstantinou you mean `/api/things/lookup?query=and&limit=4&page=1` ? It does not work.

georgeconstantinou: For me it works, do you have a record that includes `and` in the name?

asaliev:
AFAIK the API can also respond with 404 (when debug is disabled) when you request a page which is out of bounds. Not sure if you want to include this as part of the swagger response samples.

src/Swagger/Annotation.php
acrias: Okay it works. Most probably I was only changing the url instead of the actual endpoint inthe annotation file

PR #968, opened 2020-04-06
Prevent show help tooltips on empty message (task #15480)

eliamariutti:
Small bugfix on help tooltip : few csv migration field are not passing the help value to the template causing show the question mark icons without any text.

### DIFF
--- a/src/View/Helper/MyHtmlHelper.php
+++ b/src/View/Helper/MyHtmlHelper.php
@@ -21,7 +21,7 @@ class MyHtmlHelper extends HtmlHelper
         $classCss = Configure::read("CsvMigrations.HelperConfig.classCss");
         $dataPlacement = Configure::read("CsvMigrations.HelperConfig.dataPlacement");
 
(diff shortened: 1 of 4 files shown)


src/View/Helper/MyHtmlHelper.php
```diff
@@ -21,7 +21,7 @@ public function help(string $message): string
         $classCss = Configure::read("CsvMigrations.HelperConfig.classCss");
         $dataPlacement = Configure::read("CsvMigrations.HelperConfig.dataPlacement");
 
-        return '&nbsp;&nbsp;<span data-toggle="' . $dataToggle . '" class="' . $classCss . '" data-placement="' . $dataPlacement . '" data-original-title="' . __($message) . '">?</span>';
+        return empty($message) ? '' : '&nbsp;&nbsp;<span data-toggle="' . $dataToggle . '" class="' . $classCss . '" data-placement="' . $dataPlacement . '" data-original-title="' . __($message) . '">?</span>';
```
acrias: @eliamariutti shouldn't we add something to switch the tooltips on/off? An option in the settings maybe?

eliamariutti: @acrias like labels, default and other fields.json configuration, I think we just need to remove it from the file.

PR #977, opened 2020-04-23
Allow displaying non-searchable columns in search results (task #11061)

asaliev:
Allow displaying non-searchable columns in search results


src/Utility/Search.php
```diff
@@ -48,12 +48,13 @@ final class Search
...
      *
      * @param string $tableName Table name
+     * @param bool $onlySearchable Include only searchable fields
      * @return mixed[]
      */
-    public static function getFilters(string $tableName): array
+    public static function getFields(string $tableName, bool $onlySearchable = false): array
     {
         if (! empty(static::$filters[$tableName])) {
             return static::$filters[$tableName];
```
softius, line 60: @asaliev I believe you need to check for `onlySearchable` here as well since we are caching the full array. The same when checking the Cache a few lines below.

asaliev: @softius :+1:

PR #990, opened 2020-05-25
Use table alias (task #16481)

softius:
Use table alias (if defined) in system saved searches. 

The model column has been also removed as it adds no real value and presents the wrong name when the table alias is defined.


asaliev:
@softius how do you want to handle the upcoming `ModuleRegistry` changes in terms of this PR?

PR #995, opened 2020-05-27
Switch to Module classes (task #16415)

softius:
Drops calls to `ModuleConfig` and update the codebase to include `ModuleRegistry` in as many places as possible.


src/Template/Element/Module/Associated/tabs-list.ctp
```diff
@@ -16,22 +13,20 @@ $setLabels = [];
...
 
-        $label = '<span class="fa fa-' . $config->table->icon . '"></span> ';
+        $label = '<span class="fa fa-' . $config['table']['icon'] . '"></span> ';
 
         if (array_key_exists($association->getAlias(), $labels)) {
             $label .= __($labels[$association->getAlias()]);
         } else {
-            $label .= isset($config->table->alias) ?
-                __($config->table->alias) :
+            $label .= isset($config['table']['alias']) ?
```
acrias, line 23: We have many cases that we assume that keys exist without using the isset, property_exist, array_key exist. I'm not sure if we have to do those checks to avoid php warnings/errors

src/Template/Element/Module/Associated/tab-content.ctp
```diff
@@ -44,22 +42,20 @@ $dtOptions = [
             return $fields;
         }),
         'virtualColumns' => call_user_func(function () use ($module) {
-            $mc = new ModuleConfig(ConfigType::MODULE(), $module);
-            $config = $mc->parse();
+            $config = ModuleRegistry::getModule($module)->getConfig();
 
-            return property_exists($config, 'virtualFields') ? (array)$config->virtualFields : [];
+            return !empty($config['virtualFields']) ? $config['virtualFields'] : [];
```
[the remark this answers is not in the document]
softius: @acrias `empty` returns true in cases that `virtualFields` is empty or not set at all.

src/Template/Element/Module/Associated/tabs-list.ctp
softius: @acrias overall i agree that in some cases we use `empty` or `isset` for maybe the same purpose. Warnings/Errors were mostly applicable when were using objects instead of arrays. With the new setup, we shouldnt get any errors/warning as long as we don't access the value directory

PR #996, opened 2020-05-29
Switch to module classes (task #16415)

asaliev:
- Migrated templates to use `ModuleRegistry`
- Changed various helpers and formatter to use `ModuleRegistry`
- Fixed remaining broken tests


src/Template/Module/batch.ctp
```diff
@@ -11,11 +11,10 @@
...
 
 use Cake\Utility\Inflector;
-use Qobo\Utils\ModuleConfig\ConfigType;
-use Qobo\Utils\ModuleConfig\ModuleConfig;
+use Qobo\Utils\Module\ModuleRegistry;
 
-$config = (new ModuleConfig(ConfigType::MODULE(), $this->name))->parse();
-$alias = __('{0}', isset($config->table->alias) ? $config->table->alias : Inflector::humanize(Inflector::underscore($this->name)));
+$config = ModuleRegistry::getModule($this->name)->getConfig();
+$alias = __('{0}', isset($config['table']['alias']) ? $config['table']['alias'] : Inflector::humanize(Inflector::underscore($this->name)));
```
acrias, line 17: I think we should avoid using this {0} alone as it appears int the .pot file and you do not know what to translate. We are using it and in some other places in other files but do we need this single {0}. I think only __() it will be fine except the warnings/errors in the extract.

softius: @acrias i see what you saying, but we must leave that for another PR as the scope here is to replace `ModuleConfig` with `ModuleRegistry`. Its a large effort with lots of changes already :)

PR #1009, opened 2020-06-17
Allow file sorting (Task #10489)

acrias:
This PR enable file sorting for file uploads.Fileinput `theme` and some options can be changed from settings

### DIFF
--- a/src/Controller/Api/V1/V0/FileStorageController.php
+++ b/src/Controller/Api/V1/V0/FileStorageController.php
@@ -3,7 +3,6 @@
 namespace App\Controller\Api\V1\V0;
 
 use App\Service\OrderFileStorage;
-use Cake\ORM\TableRegistry;
 
 class FileStorageController extends AppController
 {
--- a/src/Controller/Api/V1/V0/FileStorageController.php
+++ b/src/Controller/Api/V1/V0/FileStorageController.php
@@ -18,7 +18,21 @@ class FileStorageController extends AppController
 
         $data = (array)$this->request->getData();
 
-        $result = OrderFileStorage::orderFiles($data);
+        try {
+            OrderFileStorage::orderFiles($data);
+
+            $result = [
+                'success' => true,
+                'data' => [],
+                'message' => __('File order has been changed'),
+            ];
+        } catch (\InvalidArgumentException | \ErrorException $e) {
+            $result = [
+                'success' => false,
+                'data' => [],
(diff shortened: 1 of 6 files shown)


src/Service/OrderFileStorage.php
```diff
@@ -0,0 +1,41 @@
...
+use Cake\ORM\TableRegistry;
+
+final class OrderFileStorage
+{
+    /**
+     * Order files
+     * @param  mixed[] $files Files to order
+     * @return mixed[] result
+     */
+    public static function orderFiles(array $files): array
```
softius, line 14: @acrias i believe if you abstract this a bit more, you could have something more generic that could be used with other Tables/Fieds and eventually become a Table Behavior.

acrias: @softius do you want to do this now or in a new ticket because the change requires time and the tests will need also refactoring

softius: @acrias i was having in mind the sequence behavior but it looks like a different problem here as there is no need for `moveUp` / `moveDown`. However, we need at least move the `$result` structure into Controller as we dont want the service to be aware of the API response. 

This can be done by adjusting `orderFiles` to be `void` and throw an `Exception` whenever something is wrong; the exception can be caught in controller and adjust the result to be returned accordingly. Tests must adjusted also, to assert that the exception is thrown.

src/Service/OrderFileStorage.php
```diff
@@ -0,0 +1,41 @@
...
+    {
+        $result = [
+            'success' => false,
+            'data' => [],
+            'message' => __('There was an error changing file order'),
+        ];
+
+        if (empty($files) || ! is_array($files)) {
```
softius, line 22: @acrias no need for `is_array` here, the argument is already type-hinted as `array`

src/Controller/Api/V1/V0/FileStorageController.php
```diff
@@ -2,6 +2,38 @@
...
+
+        try {
+            OrderFileStorage::orderFiles($data);
+
+            $result = [
+                'success' => true,
+                'data' => [],
+                'message' => __('File order has been changed'),
+            ];
+        } catch (\InvalidArgumentException | \ErrorException $e) {
```
anvyst, line 28: @acrias can you please define the Exceptions on the top of the file with `use` statement?

PR #1020, opened 2020-07-20
Lookup method does not return correct results (Task #17074)

acrias:
This PR fixes an issue where fallback to display_field didn't work in case the typeahead_fields in the table config were empty


src/Event/Controller/Api/LookupActionListener.php
```diff
@@ -259,6 +259,10 @@ protected function _getTypeaheadFields(Table $table): array
         $config = ModuleRegistry::getModule($table->getRegistryAlias())->getConfig();
         $fields = Hash::get($config, 'table.typeahead_fields', [$table->getDisplayField()]);
```
softius, line 260: @acrias lets remove the default value from here (to avoid the same default appear twice)

PR #1021, opened 2020-07-23
Fix integer on Settings index page (task #15054)

eliamariutti:
Numbers in PHP apparently are valid JSON values and it breaks the Settings
image

### DIFF
--- a/config/settings.php
+++ b/config/settings.php
@@ -57,7 +57,7 @@ return [
                     ],
                     'Port' => [
                         'alias' => 'EmailTransport.default.port',
-                        'type' => 'string',
+                        'type' => 'integer',
                         'help' => '',
                         'scope' => [
                             'app',
--- a/config/settings.php
+++ b/config/settings.php
@@ -141,7 +141,7 @@ return [
                     ],
                     'Version' => [
                         'alias' => 'Ldap.version',
-                        'type' => 'string',
+                        'type' => 'integer',
                         'help' => '',
                         'scope' => [
                             'app',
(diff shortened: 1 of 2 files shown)


src/Template/Settings/index.ctp
```diff
@@ -148,9 +148,9 @@ $(document).ready(function(){
                                                 'fieldDefinitions' => $definition,
                                                 'label' => $field
                                             ];
-                                            if (json_decode($value) && json_last_error() === JSON_ERROR_NONE) {
-                                                $fieldDefinition['attributes'] = ['readonly' => true];
-                                            }
+											if (!is_integer($value) && json_decode($value) && json_last_error() !== JSON_ERROR_NONE) {
```
acrias, line 151: @eliamariutti we want to have the port readonly? because we set the port per client. Or is not what this readonly attribute does

config/settings.php
```diff
@@ -57,7 +57,7 @@
                     ],
                     'Port' => [
                         'alias' => 'EmailTransport.default.port',
-                        'type' => 'string',
+                        'type' => 'integer',
```
softius, line 60: @eliamariutti dont we need to adjust `Version` as well?

eliamariutti: @softius is it always an integer the version ?

src/Template/Settings/index.ctp
eliamariutti: @acrias The `readonly` is enable when we have json data.

PR #1022, opened 2020-07-23
Enable Setting search (task #17130)

eliamariutti:
Enable Setting search with a small improvement 
image

### DIFF
--- a/src/Template/Settings/index.ctp
+++ b/src/Template/Settings/index.ctp
@@ -26,7 +26,8 @@ $(document).ready(function(){
 	$('#settings-search').on('input',function(){
 		let search = $(this).val();
 		let labels = $('.tab-content').find('label')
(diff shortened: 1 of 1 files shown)


src/Template/Settings/index.ctp
```diff
@@ -68,18 +70,20 @@ $(document).ready(function(){
 		<div class="col-md-12">
 			<?= $this->Form->create($settings); ?>
 			<div class="nav-tabs-custom">
-				<ul class="nav nav-tabs <?= (1 < count($data)) ? '' : 'hidden' ?>">
```
acrias, line 71: @eliamariutti I think we need this so that the whole section is not shown

eliamariutti: @acrias I need the nav-tab be always visible because in it, there is the search bar. The logic of counting data has been moved just few line after before the foreach.

PR #1025, opened 2020-07-30
Support for raw format in search and multi-field sorting (Task #17211)

acrias:
This PR allow you to get raw results from the search api and also enables you to sort the results using multiple fields


acrias:
This has been back-ported

